PR #116, opened 2015-03-16
Initial version of kano-tzupdate

skarbat:
* This script is meant to replace tzupdate to provide
   our own authentication against geonames, and use
   a free web service for the geolocation.

### DIFF
--- a/bin/kano_tzupdate
+++ b/bin/kano_tzupdate
@@ -31,7 +31,7 @@ Set the system timezone based on IP geolocation.
 # License: http://www.gnu.org/licenses/gpl-2.0.txt GNU General Public License v2
 #
 # This script has been adapted on KanoOS to provide for custom authentication
-# against geonames, and getting current gelocation from the free webservice ipinfo.io
+# against geonames, and getting current geolocation from the free webservice ipinfo.io
 #
 
 import argparse
@@ -128,7 +128,7 @@ if __name__ == "__main__":
     args = parser.parse_args()
 
     try:
-        coords = get_gelocation(ip=args.ip)
+        coords = get_geolocation(ip=args.ip)
     except LookupError as e:
         print "IPInfoDB returned an error: " + str(e)
         sys.exit(3)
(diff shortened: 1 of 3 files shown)


bin/kano_tzupdate
```diff
@@ -0,0 +1,156 @@
...
+
+        try:
+            os.unlink(PATH["localtime"])
+        except OSError as e:
+            if e.errno != errno.ENOENT:
+                raise
+
+        os.symlink(timezone_path, PATH["localtime"])
+
+    print(timezone)
```
alex5imon: do we need this print or is it debug?

alex5imon:
@Ealdwulf please review

bin/kano_tzupdate
skarbat: This print will give us the country code which is caught by the udhcp script to set the timezone under /etc/localtime, below (which needs to change to `kano_tzupdate`:

 * https://github.com/KanoComputing/kano-toolset/blob/master/udhcpc/kano.script#L95

@alex5imon @Ealdwulf

alex5imon: great, then it seems okay to me @skarbat

bin/kano_tzupdate
```diff
@@ -0,0 +1,156 @@
...
+        help="print the timezone, but don't update " + PATH["localtime"]
+    )
+    parser.add_argument(
+        "-a", "--ip",
+        help="use this IP instead of automatically detecting it"
+    )
+    args = parser.parse_args()
+
+    try:
+        coords = get_gelocation(ip=args.ip)
```
Ealdwulf: I think this should be get_geolocation

skarbat: that's right @Ealdwulf thanks!

Ealdwulf: Okay this looks good to merge to me!

alex5imon:
@skarbat please merge :+1:

skarbat:
thanks guys! @Ealdwulf @alex5imon

PR #117, opened 2015-03-18
The utils method run_bg returns the Popen object

skarbat:
* This change enables apps to run processes in the background
   and get control over their status and termination


convolu:
Nice one @skarbat

alex5imon:
Just tested.

Please merge :+1:

PR #120, opened 2015-03-25
Gtk at warning

Ealdwulf:
This PR allows the NO_AT_BRIDGE environment variable through sudo so kano-settings can be run with no warning.

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -12,7 +12,7 @@ case "$1" in
     configure)
 
         # Create custom sudoers file
-        echo "Defaults env_keep += KLOG_FORCE_FLUSH" > $TMP_FILE
+        echo 'Defaults env_keep += "KLOG_FORCE_FLUSH NO_AT_BRIDGE"' > $TMP_FILE
         echo "%sudo   ALL=(root) NOPASSWD: /usr/bin/kano-launcher" >> $TMP_FILE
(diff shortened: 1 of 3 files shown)


skarbat:
looks good to me @Ealdwulf , pretty amazing we can debug it remotely.

Ealdwulf:
Oops, I hadn't intended to make this one depend on those: I'll merge it after those are merged.

debian/postinst
```diff
@@ -12,7 +12,7 @@ case "$1" in
     configure)
 
         # Create custom sudoers file
-        echo "Defaults env_keep += KLOG_FORCE_FLUSH" > $TMP_FILE
+        echo 'Defaults env_keep += "KLOG_FORCE_FLUSH NO_AT_BRIDGE"' > $TMP_FILE
```
alex5imon: @Ealdwulf are these quotes okay? Just to double check, cause it's different from what we used before

Ealdwulf: @alex5simon, it is good to double check; in this case I had to change the
quotes because I need double quotes to end up in the resulting file. I have
checked that the file is generated correctly.

Best,
Alex

alex5imon:
Please merge :+1: 

Add to backlog

Ealdwulf:
Thanks guys!, merging...

PR #122, opened 2015-03-25
Kano wifi gui

skarbat:


alex5imon:
@skarbat @carolineclark kano-wifi-gui needs to live in `kano-settings` rather than toolset.

alex5imon:
@skarbat this PR needs to be moved to `settings` rather than `toolset`

pazdera:
Should we do something about this?

alex5imon:
yeah @pazdera we should close it :)

Thanks!

PR #123, opened 2015-03-26
Enable UTF8 for kano.logging

hzulla:
At the moment, kano.logging cannot handle unicode in log messages:

    from kano.logging import logger
    logger.force_log_level("info")
    logger.info(u"Hell\N{LATIN SMALL LETTER O WITH DIAERESIS} W\N{LATIN SMALL LETTER O WITH DIAERESIS}rld\N{HORIZONTAL ELLIPSIS}")

This patch offers a fix for that.

### DIFF
--- a/bin/kano-logs
+++ b/bin/kano-logs
@@ -88,7 +88,7 @@ def show_logs(app=None, linearised=False):
             output += format_log_data(get_log_data(log)) + "\n"
 
     if len(output) > 0:
-        pydoc.pipepager(output, cmd='less -R')
+        pydoc.pipepager(output.encode('utf-8'), cmd='less -R')
 
 
 def format_log_data(data, default_app_name=""):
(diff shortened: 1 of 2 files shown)


alex5imon:
@Ealdwulf can you review this PR please?

Ealdwulf:
I think this patch is good, @alex5imon . However when I tried looking at the resulting logs with kano-logs, it crashed in format_log_data, because it also doesn't support unicode.  Maybe we should wait to merge this until we have fixed kano-logs as well.

hzulla:
@Ealdwulf latest commit fixes kano-logs crashing with Unicode content in log files.

However, it raises the question wether one should add `.encode('utf-8')` for individual lines handling string output (unicode as the exception to the rule) or if it's smarter to switch output encoding for the python app, globally (unicode as the default encoding). What do you think?

hzulla:
Sorry, I'm retracting this PR. See https://github.com/KanoComputing/kano-updater/pull/90 for discussion why.

Ealdwulf:
Because we are only going to log in English? Okay, Thanks anyway.

hzulla:
No, only logging English by policy isn't enough. Yeah, logs should be English, but even if all log entries were English, some log content will be non-ASCII. (E.g., what happens when a user's name is "Jürgen"?) So kano-toolset needs to be unicode-safe as a whole.

But adding .encode() to each line handling string i/o, as I did for this PR, isn't elegant. It's error prone and in the long run, it will result in conversion hell. Unless you have a plan to switch to Python 3 in the near future, I guess one needs a more general approach to switch string and file i/o to utf8, so that one doesn't have to add .encode() every time.

Ealdwulf:
Okay, that makes sense;  thanks for the explanation

PR #124, opened 2015-04-01
Fixing kano-launcher when no configuration available

skarbat:
* Set defaults for when no configuration file is found

cc @Ealdwulf


skarbat:
like so @Ealdwulf ?

Ealdwulf:
Yes, thanks Albert!

skarbat:
Thanks @Ealdwulf merging!

PR #125, opened 2015-04-08
Add functions for minecraft improvments.

Ealdwulf:
This PR adds functions for use in minecraft: 

* kano.windows.get_child_windows: for finding the MPi child window so we can resize it 
* kano.xwindow: functions for use by the daemon which catches tabs

Because make-minecraft will depend on these changes, the version is also incremented.

### DIFF
--- a/kano/window.py
+++ b/kano/window.py
@@ -14,6 +14,7 @@ from gtk import gdk
 import time
 
 from kano.utils import run_cmd
+import kano.xwindow
 
 BOTTOM_BAR_HEIGHT = 44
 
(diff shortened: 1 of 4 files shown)


kano/window.py
```diff
@@ -134,6 +135,19 @@ def _get_window_by_id(wid):
...
+def retry_for(fn, secs, *args, **kwargs):
+    '''
+    Retry a function for some time if it returns None
+    '''
+    res = None
+    for i in range(1, secs * 10):
+        res = fn(*args, **kwargs)
+        if res is not None:
+            return res
+        time.sleep(0.1)
```
tombettany: Only concern is that the sleep time here (0.1) is highly dependant on the factor for the range, i.e. the 10 in `for i in range(1, secs * 10):`. Not really a problem but just a note.

tombettany:
@Ealdwulf Nice use of the context manager for the display. Other than the small niggle noted above this looks fine to me.

Ealdwulf:
Thanks @tombettany . I have broken the constant out and will merge.

PR #126, opened 2015-04-09
Add option to prevent the window from being minimised 

Ealdwulf:
@pazdera @tombettany 
This PR adds an option to webapp to prevent the window from being minimised,
which is useful because being minimised is confusing in minecraft. The method is the same used in kano-feedback (but translated to pygtk).


tombettany:
I don't see any problems here.

Ealdwulf:
Thanks, merging...

PR #127, opened 2015-04-15
Execute DHCP network hooks only for when a new lease is acquired

skarbat:
Because the network hook currently has several Python modules, and it is only needed during a new lease, we are removing these extra hooks to avoid an unnecessary performance penalty.


pazdera:
Looks good to me! :shipit:

Feel free to merge!

PR #128, opened 2015-04-17
Add a sequence number to  backend-api.js 

Ealdwulf:
This PR should fix a problem in minecraft whereby tab from blocks mode stops working. It was caused by the fact that the API call is encoded as a hash, and the signal  for the hash only gets triggered when the hash string changes, so if the same function is called with the same arguments it doesn't happen. Adding a sequence number to make sure there is some change.

@tombettany @pazdera

### DIFF
--- a/js/backend-api.js
+++ b/js/backend-api.js
@@ -11,6 +11,7 @@
 
 window.backend = {
     callbacks: [],
+    seq: 0,
 
     call: function() {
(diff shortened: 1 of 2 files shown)


pazdera:
Looks good! :-)

Please merge :shipit:

Ealdwulf:
Thanks! Merging...

PR #129, opened 2015-04-28
Key Signals: Allow propagation of key presses

tombettany:
KanoComputing/peldins#1419
Currently, `xbindkeys` can only propagate commands to Pong and Minecraft
and absorb everything else. Instead, catch these Pong and Minecraft
events and trigger them in the apps, if not, instead sane propagate key
presses to the app to allow handling further down.

cc @skarbat


skarbat:
@tombettany this looks good to me.

PR #131, opened 2015-05-11
Stop the system reconnecting to internet after actively disconnecting.

carolineclark:


alex5imon:
@carolineclark is `disconnect` function used anywhere else? Do we always want to clean the cache?

carolineclark:
@alex5imon disconnect doesn't seem to be used elsewhere, and I don't need to clear the cache for it to disconnect, I presumed this would stop it connecting after a reboot, but I didn't test.

alex5imon:
If it's not used anywhere else then cleaning the cache is okay.

Please merge :+1: 

Add to changelog

carolineclark:
Whoops, I'm wrong, it's used here https://github.com/KanoComputing/kano-settings/blob/9c3dbe6918e5e30f7a804f0c1b7875d768f4bc85/bin/kano-connect#L164

alex5imon:
@carolineclark in this case you might not want to clear the cache. So maybe so can pass an optional  argument (default to False). The call it from your new functionality disconnect(iwface, True)

PR #132, opened 2015-05-12
Network: Make network_info report unaltered ESSID

tombettany:
KanoComputing/peldins#1832
Currently, to report the ESSID, `iwconfig` is queried and the output is
passed through grep, awk and sed. During this process, everything after
a space is scrapped and some of the other characters are lost. Instead
use Python's regex module to handle this to retain the spaces and
special characters.

cc @pazdera @convolu


pazdera:
Looks great! Please merge :shipit:.

convolu:
Looks good to me :+1:

PR #133, opened 2015-05-13
utils: Adding get_free_space

pazdera:
A function to check for the amount of free space on disk.

This will be useful in the updater for checking how much space we have before updating.

Related to: https://github.com/KanoComputing/peldins/issues/1798

cc @alex5imon @tombettany


alex5imon:
Looks good to me! 

Please merge :+1:

PR #134, opened 2015-05-13
Allow for graceful exit for chown

convolu:
Avoid crashing if chown is accidentally called without sudo privileges.

### DIFF
--- a/kano/utils.py
+++ b/kano/utils.py
@@ -293,9 +293,23 @@ def chown_path(path, user=None, group=None):
         user = user_unsudoed
     if not group:
         group = user_unsudoed
-    uid = pwd.getpwnam(user).pw_uid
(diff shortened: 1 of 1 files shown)


convolu:
@alex5imon @pazdera @skarbat Could you please review?

alex5imon:
Seems okay to me. My only comment is that we are now returning a value, just take that into account where this function is called in other parts of the system

convolu:
Since this is performing a fairly important function it would be useful to check whether it has failed since the may prevent the program calling it from continuing.

I have greped through our code and there doesn't seem to be anywhere where the added return value is affecting the code execution.

skarbat:
it looks good to me @convolu

convolu:
Merged, thanks :smile:

PR #136, opened 2015-05-15
Kano Signal: Communicate to Sonic Pi through dbus

tombettany:
KanoComputing/peldins#1419
Kano signal used to use `xdotool` to propagate the Save, Share, Load and
Make keys to the application. Instead switch to use dbus.

cc @convolu @skarbat

### DIFF
--- a/bin/kano-signal
+++ b/bin/kano-signal
@@ -22,19 +22,15 @@ signal=$1
 # syntax sanity check
 case $signal in
     save)
-        key_propagation="ctrl+s"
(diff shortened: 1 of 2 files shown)


skarbat:
magic! looks good to me @tombettany

convolu:
Looks good to me :shipit:

tombettany:
@Ealdwulf, yeah `kano-signal` uses `xwininfo` but that was missing.

PR #137, opened 2015-05-15
Add unsudo option

Ealdwulf:
This PR adds an 'unsudo' option to run_cmd, run_cmd_log, run_cmd_bg to allow running as the
original user. This is to avoid running things unnecessrily as root which can result in files
being owned as root.

It also bumps the revision since this will need to be used in other packages.

### DIFF
--- a/kano/utils.py
+++ b/kano/utils.py
@@ -22,7 +22,7 @@ import grp
 import json
 
 
-def run_cmd(cmd, localised=False):
+def run_cmd(cmd, localised=False, unsudo=False):
     '''
     Executes cmd, returning stdout, stderr, return code
     if localised is False, LC_ALL will be set to "C"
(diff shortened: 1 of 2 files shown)


kano/utils.py
```diff
@@ -31,6 +31,9 @@ def run_cmd(cmd, localised=False):
     if not localised:
         env['LC_ALL'] = 'C'
 
+    if unsudo and 'SUDO_USER' in os.environ and os.environ['SUDO_USER'] != 'root':
```
tombettany: We actually already have a function in this file ([`get_user_unsudoed()`](https://github.com/KanoComputing/kano-toolset/blob/d12e1b60c6c0e852f41932561f3043bf311a961d/kano/utils.py#L173-L179)) for retrieving the unsudo'd user - maybe we should use that instead?

kano/utils.py
```diff
@@ -69,6 +72,9 @@ def run_bg(cmd, localised=False):
...
+        cmd = "sudo -u {} bash -c '{}' ".format(os.environ['SUDO_USER'], cmd)
```
tombettany: Likewise here.

Ealdwulf: @tombettany The logic is not quite the same as get_user_unsudoed(), because
we don't want to sudo for each run_cmd, only those for which we are
actually root. It may be that some logic can be shared though (if nothing
else, between the two locations you commented on).

alex5imon:
@Ealdwulf @tombettany do we want to merge this?

tombettany:
@Ealdwulf I think that you have allayed my only objections to this so I'm happy to proceed.

Ealdwulf:
Thanks @tombettany , merging...

PR #138, opened 2015-06-04
i18n: Add module for handling i18n setup

tombettany:
A module which can import and install gettext (including defining the
function for deferred translations).

Depends on https://github.com/KanoComputing/kano-toolset/pull/139


tombettany:
This got moved to https://github.com/KanoComputing/kano-i18n so closing here.

PR #139, opened 2015-06-04
Add module for app initialisation

tombettany:
This module replaces the standard calls at the top of every executable
to add the local path to the list of directories for Python to search
through with:

 ```python
    from kano.app import init_env
    init_env('app_name', __file__)
```

cc @pazdera @convolu @Ealdwulf @carolineclark @alex5imon

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,9 @@
+kano-toolset (2.0-2) unstable; urgency=low
+
+  * Add app module
+
+ -- Team Kano <dev@kano.me>  Thu, 4 Jun 2015 16:00:00 +0100
+
 kano-toolset (2.0-1) unstable; urgency=low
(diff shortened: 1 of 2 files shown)


convolu:
Maybe we need a better name :-P.

convolu:
Looks good to me

tombettany:
@pazdera @convolu @Ealdwulf @carolineclark @alex5imon @skarbat  This is the PR for replacing the executable header that we were discussing this morning. Any thoughts, improvements or alternative plans to remove this burden that we carry with every executable?

pazdera:
Looks good to me :-). :shipit: 

The example in the PR says `from kano.apps` which would be pretty confusing, but I see that the module in code is `.app`.

PR #140, opened 2015-06-22
Adding support for Internet status via a regular text file

skarbat:
* Script changes for UDHCP hook, and new ifplugd script are provided,
   which will maintain an external regular file to monitor the internet status
 * The file can then be monitored for changes as a trigger (lxpanel status, etc)

### DIFF
--- a/bin/kano-network-hook
+++ b/bin/kano-network-hook
@@ -17,6 +17,9 @@ import time
 from kano.logging import logger
 from kano.utils import get_cpu_id, get_rpi_model, run_cmd
 
+# file that external applications can monitor for internet availability triggers
+monitor_file="/var/opt/internet_monitor"
+
 def send_cpu_id():
     sent=False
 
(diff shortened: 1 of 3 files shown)


skarbat:
@alex5imon @tombettany does that look good to you ?

alex5imon:
@pazdera is this the best way of doing this, what do you think? https://github.com/KanoComputing/peldins/issues/1860

pazdera:
Yeah, it's one of the two ways I see here:

* create a monitor file that Albert did here and use the glibs file monitor like the updater widget
* create a pipe from the widget itself and pipe any events that occur to it through there (this is what the notification widget does)

With lxpanel widgets its tricky, because you can't really signal them, they have to be polling something on itheir own.

In this case, Albert chose the better one, because if we want to hook more tools to the network event, they can just look at the same file. With the other approach that wouldn't be possible. So for me :+1:.

skarbat:
Thanks for looking into this Radek! @pazdera

PR #141, opened 2015-06-23
Gtk: Remove unnecessary padding from LabelledEntries

tombettany:
KanoComputing/peldins#1883
The LabelledEntries class introduced a margin which maligns any centring
of the widget.

cc @pazdera @carolineclark @convolu

### DIFF
--- a/kano/gtk3/labelled_entries.py
+++ b/kano/gtk3/labelled_entries.py
@@ -14,7 +14,6 @@ class LabelledEntries(Gtk.Alignment):
 
     def __init__(self, entries_info):
         Gtk.Alignment.__init__(self)
-        self.set_padding(0, 0, 0, 100)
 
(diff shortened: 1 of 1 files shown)


pazdera:
:shipit: :tractor:

convolu:
From a quick grep this seems to be used in 
* https://github.com/KanoComputing/kano-settings/blob/57048e31fb21a0ddce0f447e4fe6a5f183755685/kano_settings/set_account.py#L195
* https://github.com/KanoComputing/kano-profile/blob/00e4b25a15c879a480bb41d394c62bdd590d6b3c/kano_login/login.py#L59

Thus we shouldn't expect anything breaking because of it.

Looks good :+1:

tombettany:
@convolu I modified this for `kano-profile` and checked it for `kano-settings` so I think that we should be okay.

PR #142, opened 2015-06-25
Midori

skarbat:
Provide new function to open the default system browser which will now point to Midori.


skarbat:
cc @alex5imon @rluz

alex5imon:
Please merge :+1:

PR #143, opened 2015-06-26
Logging: Fix flush and sync functions

tombettany:
When `flush()`ing the log, it would close the file if one existed,
rather than flushing the file. It would also flush the `stderr`.

`sync`, on the other hand, would `flush()` the log file regardless of
whether it was open but would not flush the `stderr`.

Merge these two functions but retain both interfaces, ensuring that the
log file is open before flushing, flush the `stderr` and don't close the
file in any circumstance.

cc @pazdera

### DIFF
--- a/kano/logging.py
+++ b/kano/logging.py
@@ -197,7 +197,7 @@ class Logger:
                     sys.stderr.write(output_line)
 
     def sync(self):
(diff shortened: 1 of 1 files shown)


pazdera:
Looks good, :shipit:

PR #144, opened 2015-07-01
Timeout

pazdera:
Adding the timeout decorator to be used from `notifications.py`.

If there isn't anything at the other end of the pipe, writing into it will block forever. This will be helpful to fix that.

cc @alex5imon @tombettany

### DIFF
--- a/kano/timeout.py
+++ b/kano/timeout.py
@@ -1,5 +1,8 @@
 # The timeout decorator
 #
+# Copyright (C) 2015 Kano Computing Ltd.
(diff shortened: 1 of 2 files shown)


kano/timeout.py
```diff
@@ -0,0 +1,40 @@
+# The timeout decorator
+#
```
alex5imon: needs copyright header @pazdera (even with the source)

pazdera:
@alex5imon Updated the header

alex5imon:
Please merge :+1:

PR #147, opened 2015-08-12
Fix for Realtek Wifi problems

darryllee:
Per discussion here:  http://world.kano.me/forum/topic/wifi-not-working-in-2.0/

* Driver parameter for wpa_supplicant can take multiple drivers, so adding wext in addition to nl80211.
* After adding wext driver, timeout was still too short, so GUI kept spinning thinking association failed, while connect toolbar widget did recognize the connection (as did ifconfig).
* Lastly -- minor typo, lenght -> length


alex5imon:
@darryllee thanks a lot for the PR, sorry it took as so long to reply. 

The problem has been solved here: https://github.com/KanoComputing/kano-toolset/pull/148

PR #150, opened 2015-09-25
Adding the new cpuid tracker to the network hook

pazdera:
Depends on: https://github.com/KanoComputing/kano-profile/pull/329

### DIFF
--- a/bin/kano-network-hook
+++ b/bin/kano-network-hook
@@ -32,8 +32,11 @@ def send_cpu_id():
         try:
             # This is a loose dependency as we can't make kano-toolset
             # depend on profile.
-            from kano_profile.tracker_events import generate_event
-            generate_event('ping')
+            try:
+                from kano_profile.tracker_events import generate_event
+                generate_event('ping')
+            except Exception as e:
+                logger.error("Error sending the ping event ({})".format(e))
 
             # The code below is the legacy way of tracking cpuids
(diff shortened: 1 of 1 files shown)


pazdera:
Sorry guys, I forgot to tag you, @alex5imon and @skarbat.

skarbat:
looks great @pazdera . I can't see the dependency towards kano-profile, so perhaps would be good to embrace this task in an exception block, and pass on failure, so the rest of the code proceeds?

pazdera:
That's a good idea. I'll put it in an additional try.

The reason while we don't have the dependency there is because it would be
circular. Profile depends on toolset :-/.



On Monday, 28 September 2015, Albert Casals <notifications@github.com>
wrote:

> looks great @pazdera <https://github.com/pazdera> . I can't see the
> dependency towards kano-profile, so perhaps would be good to embrace this
> task in an exception block, and pass on failure, so the rest of the code
> proceeds?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/KanoComputing/kano-toolset/pull/150#issuecomment-143775962>
> .
>


-- 
Radek Pazdera
Software Engineer at Kano <http://www.kano.me>

pazdera:
@skarbat Fixed, thanks!

skarbat:
great one @pazdera Radek!

pazdera:
Cheers, @skarbat! I'll merge this.

PR #152, opened 2015-10-14
Volume jessie fix

rluz:
Control Master volume only for Jessie, no need for PCM control as
Chromium does not require jackd for audio.

@alex5imon @convolu

### DIFF
--- a/bin/kano-volume
+++ b/bin/kano-volume
@@ -27,21 +27,15 @@ fi
 # First parameter is either "up" or "down"
 updown=$1
 
-# percentage to increase / decrease the volume
-up_factor=8
-down_factor=7
-
 case $updown in
     "up")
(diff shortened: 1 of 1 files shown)


alex5imon:
how does this affect non-jessie users? @rluz

convolu:
@rluz You may want to rebase the jessie branch on top of master and then cherry pick your commit over a new branch because if we merge this PR we will have an unnecessarily complicated history.

@alex5imon This is a Jessie-only change

pazdera:
Rebased here: https://github.com/KanoComputing/kano-toolset/pull/153

rluz:
@alex5imon it doesn't affect wheezy user because they will still have pcm linked to master volume and midori for youtube and jessie users will have only Master volume and chromium for youtube

PR #155, opened 2015-10-15
Set Chromium as default Internet browser

rluz:
Set Chromium as default Internet browser

@pazdera @alex5imon


pazdera:
Looks good!

PR #156, opened 2015-10-16
Send the ping event on every boot

pazdera:
Part of the unreliability of the process is that we only send it once. So if the server rejects it for some reason (a bug in the API) we will have lost the event forever. This changes it to happen on every connection to a network.

cc @alex5imon @skarbat


skarbat:
that looks good to me @pazdera Radek

PR #158, opened 2015-10-21
Revert symlink

pazdera:
This will now be handled by the chromium shell package until we resolve the dependency on it.

cc @alex5imon @skarbat


skarbat:
Looks good to me @pazdera , much better.

PR #159, opened 2015-10-22
Add open_locked() from kano-profile as it is generally useful.

Ealdwulf:
This PR adds @pazdera's `open_locked` from kano-profile.tracker to kano-toolset, as I want to use it in kano-settings.boot_config
I know, andother api in kano.utils. However, we need to split it anyway and I'm not about to do so at this point.
@pazdera @tombettany 
Will rebase jessie branch once this is merged.

### DIFF
--- a/kano/utils.py
+++ b/kano/utils.py
@@ -584,3 +584,13 @@ def get_free_space(path="/"):
     device, size, used, free, percent, mp = out.split('\n')[1].split()
 
     return int(free) / 1024
+
+
+class open_locked(file):
+    """ A version of open with an exclusive lock to be used within
+        controlled execution statements.
+    """
+    def __init__(self, *args, **kwargs):
+        super(open_locked, self).__init__(*args, **kwargs)
+        fcntl.flock(self, fcntl.LOCK_EX)
+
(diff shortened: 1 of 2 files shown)


convolu:
Looks good to me :+1:

kano/utils.py
```diff
@@ -584,3 +584,13 @@ def get_free_space(path="/"):
...
     return int(free) / 1024
+
+
+class open_locked(file):
+    """ A version of open with an exclusive lock to be used within
+        controlled execution statements.
+    """
+    def __init__(self, *args, **kwargs):
+        super(open_locked, self).__init__(*args, **kwargs)
+        fcntl.flock(self, fcntl.LOCK_EX)
```
tombettany: Do we want to expose the ability to run non-blocking with `fcntl.LOCK_NB`?

convolu: That is a good idea. We could add it as an optional argument.

We don't seem to use it as a mechanism anywhere at the moment, i.e. we expect all of the `open` s to succeed, but we may use it in the future.

Ealdwulf:
Okay @convolu  @tombettany , I've added a nonblocking option.
Can I merge this?

convolu:
:shipit:

PR #161, opened 2015-10-28
ifplugd: Flush IP addresses when connection is lost

tombettany:
When a network interface goes down, the IP addresses associated with
that interface are retained. This leads to troubles for applications
which use the IP address to determine connectivity. Resolve this by
flushing all the associated IP addresses when the interface is detected
as down.

@convolu @Ealdwulf @skarbat @pazdera Do you see any ramifications
from doing this?

@alex5imon This resolves the Make Art long waiting time problem.

### DIFF
--- a/ifplugd/kano-ifplugd
+++ b/ifplugd/kano-ifplugd
@@ -17,6 +17,7 @@
 
 # file that external applications can monitor for internet availability triggers
 monitor_file="/var/opt/internet_monitor"
+interface="$1"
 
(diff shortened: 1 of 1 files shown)


convolu:
This fixes the long delay problem we saw in make-art where for some reason the requests were sent to localhost only after the timeout from the interface that hasn't been flushed.

skarbat:
it seems to work great, good finding @tombettany

PR #162, opened 2015-10-28
Replaced shutdown/reboot with systemctl

radujipa:
related to https://github.com/KanoComputing/peldins/issues/2122


radujipa:
@skarbat could you please review?

PR #163, opened 2015-10-29
Disable udev ifplugd

Ealdwulf:
@skarbat This is the cange we discussed


skarbat:
looks perfect to me @Ealdwulf

PR #165, opened 2015-12-24
Utils: Separate into submodules

tombettany:
Split utility functions into several files rather than one very large
(and ever-expanding) file. Import all the functions with `util`'s
`__init__.py` to provide backwards compatibility.

Thoughts @convolu @Ealdwulf?


Ealdwulf:
Really neat @tombettany :+1:

convolu:
long overdue. :shipit: 

It may be worth informing everyone once this is merged just in case we start getting weird issues.
cc @KanoComputing/software-team

tombettany:
Should this actually be merged to `jessie` rather than `master`?

convolu:
I would go for `jessie` to keep it consistent

PR #166, opened 2015-12-24
Decorators: Add module for useful decorators

tombettany:
Decorators are very powerful and there are often occasions where a
decorator can be reused by multiple packages. Add a first decorator for
`require_root` to the module.

Taken out of https://github.com/KanoComputing/make-light/pull/180.


convolu:
+1

convolu:
It would also be useful to add another one that on failure calls the `sys.exit()` call for our executables that can't work without sudo at all :-)

pazdera:
Nice approach!

PR #167, opened 2016-01-06
Decorators: Add module for useful decorators

tombettany:
Decorators are very powerful and there are often occasions where a
decorator can be reused by multiple packages. Add a first decorator for
`require_root` to the module.

Taken out of KanoComputing/make-light#180.

Supersedes https://github.com/KanoComputing/kano-toolset/pull/166

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,9 @@
+kano-toolset (2.2.0-3) unstable; urgency=low
+
+  * Add decorators module.
+
+ -- Team Kano <dev@kano.me>  Wed, 6 Jan 2016 16:17:00 +0100
+
 kano-toolset (2.2.0-2) unstable; urgency=low
 
   * Adding get_partition_info API.
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,4 @@
-kano-toolset (2.2.0-3) unstable; urgency=low
+kano-toolset (2.3.0-1) unstable; urgency=low
 
   * Add decorators module.
 
--- a/kano/decorators.py
+++ b/kano/decorators.py
@@ -37,7 +37,7 @@ def require_root(exit_on_failure=False, verbose=False):
                 logger.error(msg)
 
                 if verbose:
-                    print msg
+                    sys.stdout.write('{}\n'.format(msg))
 
                 if exit_on_failure:
                     sys.exit(ERR_ROOT_PERMISSIONS_REQ)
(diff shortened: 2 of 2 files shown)


tombettany:
@convolu Could you review the modified decorator version with your suggested change?

debian/changelog
```diff
@@ -1,3 +1,9 @@
+kano-toolset (2.2.0-3) unstable; urgency=low
```
convolu: Since this will be released in 2.3.0, I would suggest to bump it a bit more :-)

kano/decorators.py
```diff
@@ -0,0 +1,51 @@
...
+
+ERR_ROOT_PERMISSIONS_REQ = -1
+
+
+def require_root(exit_on_failure=False, verbose=False):
+    '''
+    Generates decorator to enforce root permissions
+
+    NB: must be called when used as decorator, i.e.
+        @require_root()
```
convolu: This may be a silly question, but are the parentheses necessary when used as a decorator?

kano/decorators.py
```diff
@@ -0,0 +1,51 @@
...
+        Actual decorator that gets applied to functions
+        '''
+
+        def ensure_root(*args, **kwargs):
+            if os.getuid() != 0:
+                msg = 'You need to run this option as root, try with sudo'
+                logger.error(msg)
+
+                if verbose:
+                    print msg
```
convolu: Is there any cases where using `sys.stdout.write(msg + '\n')` or sth equivalent instead of `print` may be better?

convolu:
Very minor points, other than that looks great :shipit:

kano/decorators.py
tombettany: For this one, yes - the `require_root` function is actually a decorator generator, not a decorator itself so needs to be called with parentheses. This is explained better in [this Stack Overflow thread](http://stackoverflow.com/questions/5929107/python-decorators-with-parameters#25827070).

debian/changelog
tombettany: Good point, I'll get on it.

kano/decorators.py
tombettany: Maybe. It is probably safer so I'll switch it anyway.

tombettany:
I'll merge then :)

kano/decorators.py
convolu: Thanks for clearing this up :-)

PR #168, opened 2016-01-06
Utils: Separate into submodules

tombettany:
Split utility functions into several files rather than one very large
(and ever-expanding) file. Import all the functions with `util`'s
`__init__.py` to provide backwards compatibility.

Supersedes https://github.com/KanoComputing/kano-toolset/pull/165


tombettany:
If nobody has any objections I'd like to get this in ASAP so that we can both test it and to prevent changes to `utils.py` from having to be moved during a rebase.

convolu:
Go go go

tombettany:
Going going going

PR #169, opened 2016-01-11
Utils: Fix circular dependency caused by separation

tombettany:
When separating Kano Utils into submodules, a circular dependency
between the `shell.py` and `process.py` modules was created. Resolve
this by migrating the offending code to be contained in just the `shell`
module.

Thanks to @Ealdwulf for pointing this out. Could you take a look?

### DIFF
--- a/kano/utils/__init__.py
+++ b/kano/utils/__init__.py
@@ -23,11 +23,12 @@ from kano.utils.hardware import \
     detect_kano_keyboard, is_model_a, is_model_b, is_model_b_plus, \
     is_model_2_b, get_rpi_model, is_monitor, get_mac_address, get_cpu_id
 from kano.utils.processes import \
-    kill_child_processes, is_running, get_program_name, pkill, restore_signals
+    kill_child_processes, is_running, get_program_name, pkill
 from kano.utils.gui import \
     is_gui, zenity_show_progress
 from kano.utils.shell import \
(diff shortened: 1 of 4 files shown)


Ealdwulf:
I think you also need to change `kano/utils/__init__.py`? Otherwise looks good.

tombettany:
@Ealdwulf You were absolutely right. Should be fixed now.

Ealdwulf:
@tombettany Looks good...

Ealdwulf:
@tombettany Actually I think you need to modify this line too: https://github.com/KanoComputing/kano-toolset/blob/jessie/kano/utils/gui.py#L4

tombettany:
@Ealdwulf Gahhhh, I'm not very good at this, am I?

tombettany:
Right @Ealdwulf, I think that I've got all of them:

```bash
$ grep -r restore_signals ./kano-toolset/

./kano-toolset//kano/utils/__init__.py:    restore_signals
./kano-toolset//kano/utils/gui.py:from kano.utils.shell import restore_signals
./kano-toolset//kano/utils/gui.py:            preexec_fn=restore_signals
./kano-toolset//kano/utils/shell.py:def restore_signals():
./kano-toolset//kano/utils/shell.py:                               preexec_fn=restore_signals)
```

Ealdwulf:
Okay @tombettany I think it's good now :+1:

tombettany:
I'm going to go for it. Wish me luck :four_leaf_clover:

PR #170, opened 2016-01-12
Fixing is_systemd function

skarbat:
* A missing import was generating an exception which made
   is_systemd() to always return False.

cc @pazdera @Ealdwulf


convolu:
Good catch @skarbat

tombettany:
Nice one @skarbat

pazdera:
:shipit: Please merge!

PR #171, opened 2016-01-14
Amixer output parsing fix

radujipa:
related to https://github.com/KanoComputing/peldins/issues/2201

### DIFF
--- a/kano/utils/audio.py
+++ b/kano/utils/audio.py
@@ -1,3 +1,10 @@
+# audio.py
+#
+# Copyright (C) 2014-2016 Kano Computing Ltd.
+# License: http://www.gnu.org/licenses/gpl-2.0.txt GNU GPL v2
+#
+
+
 import os
 
 from kano.utils.shell import run_cmd, run_bg, run_cmd_log
--- a/kano/utils/audio.py
+++ b/kano/utils/audio.py
@@ -69,26 +69,15 @@ def get_volume():
     from kano.logging import logger
 
     percent = 100
-    millibel = 400
 
-    cmd = "amixer | grep %"
-    o, _, _ = run_cmd(cmd)
-    o = o.strip().split(' ')
+    cmd = "amixer | head -n 6 | grep -Po '(\d{1,3})(?=%)'"
+    output, _, _ = run_cmd(cmd)
 
     try:
-        millibel = int(o[2])
+        percent = int(output)
     except Exception:
-        msg = 'asmixer format bad for millibel, o: {}'.format(o)
+        msg = 'amixer format bad for percent, output: {}'.format(output)
         logger.error(msg)
         pass
 
-    try:
-        percent = int(o[3].translate(None, '[]%'))
-    except Exception:
-        msg = 'asmixer format bad for percent, o: {}'.format(o)
-        logger.error(msg)
-        pass
-
-    # logger.debug('percent: {}, millibel: {}'.format(percent, millibel))
-
-    return percent, millibel
+    return percent
--- a/kano/utils/audio.py
+++ b/kano/utils/audio.py
@@ -3,6 +3,7 @@
 # Copyright (C) 2014-2016 Kano Computing Ltd.
 # License: http://www.gnu.org/licenses/gpl-2.0.txt GNU GPL v2
 #
+# Utilities relating to audio
 
 
 import os
--- a/kano/utils/audio.py
+++ b/kano/utils/audio.py
@@ -74,7 +74,7 @@ def get_volume():
     output, _, _ = run_cmd(cmd)
 
     try:
-        percent = int(output)
+        percent = int(output.strip())
     except Exception:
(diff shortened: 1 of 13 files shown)


radujipa:
@tom @convolu could you review please?

kano/utils/audio.py
```diff
@@ -62,26 +69,15 @@ def get_volume():
...
     except Exception:
-        msg = 'asmixer format bad for percent, o: {}'.format(o)
+        msg = 'amixer format bad for percent, output: {}'.format(output)
         logger.error(msg)
         pass
 
-    # logger.debug('percent: {}, millibel: {}'.format(percent, millibel))
-
-    return percent, millibel
+    return percent
```
tombettany: Should this still return a tuple? Won't some things get confused?

kano/utils/audio.py
```diff
@@ -1,3 +1,10 @@
+# audio.py
+#
+# Copyright (C) 2014-2016 Kano Computing Ltd.
+# License: http://www.gnu.org/licenses/gpl-2.0.txt GNU GPL v2
+#
```
convolu: Could you add a short description here, please?

kano/utils/disk.py
```diff
@@ -1,3 +1,10 @@
+# disk.py
+#
+# Copyright (C) 2014-2016 Kano Computing Ltd.
+# License: http://www.gnu.org/licenses/gpl-2.0.txt GNU GPL v2
+#
```
convolu: Here as well

kano/utils/file_operations.py
```diff
@@ -1,3 +1,10 @@
+# file_operations.py
+#
+# Copyright (C) 2014-2016 Kano Computing Ltd.
+# License: http://www.gnu.org/licenses/gpl-2.0.txt GNU GPL v2
+#
```
convolu: Here too

kano/utils/gui.py
```diff
@@ -1,3 +1,10 @@
+# gui.py
+#
+# Copyright (C) 2014-2016 Kano Computing Ltd.
+# License: http://www.gnu.org/licenses/gpl-2.0.txt GNU GPL v2
+#
```
convolu: Same :-(

kano/utils/audio.py
```diff
@@ -62,26 +69,15 @@ def get_volume():
...
-    except Exception:
-        msg = 'asmixer format bad for millibel, o: {}'.format(o)
-        logger.error(msg)
-        pass
+    cmd = "amixer | head -n 6 | grep -Po '(\d{1,3})(?=%)'"
+    output, _, _ = run_cmd(cmd)
 
     try:
-        percent = int(o[3].translate(None, '[]%'))
+        percent = int(output)
```
convolu: Isn't it necessary to strip `output` before you use it?

kano/utils/audio.py
radujipa: As far as I've found, only `kano-video-cli` is using this and I will update that as well soon.

convolu: There is a different helper fn that allows translation between the two, for the apps that need it https://github.com/KanoComputing/kano-toolset/blob/jessie/kano/utils/audio.py#L36.

If you go ahead with this interface change though, you should probably bump this package's version and then depend in the packages that are affected

kano/utils/audio.py
radujipa: Better safe than sorry! I'll add that in, thanks.

kano/utils/audio.py
radujipa: Updated all of them

kano/utils/audio.py
```diff
@@ -62,26 +69,15 @@ def get_volume():
...
-    o, _, _ = run_cmd(cmd)
-    o = o.strip().split(' ')
-
-    try:
-        millibel = int(o[2])
-    except Exception:
-        msg = 'asmixer format bad for millibel, o: {}'.format(o)
-        logger.error(msg)
-        pass
+    cmd = "amixer | head -n 6 | grep -Po '(\d{1,3})(?=%)'"
```
tombettany: Are we certain that it will always be the first control? Would it not be better to search for the amixer control that we want and then select the subsequent lines?

radujipa: I would say that for the time being, this is sufficient. The `Master` channel seems to always be reported first. We can have tests that look out for this if it ever fails. And if we ever add other channels or rename `Master`, chances are that there will be more changes than this.

If you'd still like a more complicated approach, I could have a look.

kano/utils/audio.py
convolu: Thanks :)

kano/utils/audio.py
convolu: Adding a test would be very useful, this was probably broken after the jesse migration and nobody noticed.

radujipa: cc @skarbat

PR #173, opened 2016-01-15
Add an optional 'timeout' parameter to open_locked(). 

Ealdwulf:
This will be used in locking the config.txt to avoid getting stuck. @pazdera


pazdera:
Looks good, @Ealdwulf! I'd only move the `errno` and `time` imports at the top of the file. Or is there a reason why you made them local?

Ealdwulf:
Thanks @pazdera . Have moved the imports - you're right, there's no reason for them to be local.

PR #175, opened 2016-01-15
Add multiline entry

convolu:
This adds the Multiline Entry widget to the toolset.

This widget was developed for Make Light. It is similar to a Textview, but has such features as:
* Kano styling
* Default placeholder text
* Maximum character limit

### DIFF
--- a/kano/gtk3/multiline_entry.py
+++ b/kano/gtk3/multiline_entry.py
@@ -1,0 +1,182 @@
+# multiline_entry.py
+#
+# Copyright (C) 2015-2016 Kano Computing Ltd.
+# License: http://www.gnu.org/licenses/gpl-2.0.txt GNU GPL v2
+#
+# Modified Gtk.TextView to permit placeholder text and other functions to make
+# it more consistent with GtkEntry
+
+
+import os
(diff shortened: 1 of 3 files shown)


convolu:
@tombettany @pazdera @radujipa Could you please review?

kano/gtk3/multiline_entry.py
```diff
@@ -0,0 +1,185 @@
...
+    @property
+    def max_length(self):
+        return self._max_length
+
+    @max_length.setter
+    def max_length(self, value):
+        if value <= 0:
+            self._max_length = 0
+        else:
+            self._max_length = value
```
tombettany: Is this not:

```python
self._max_length = max(value, 0)
```

convolu: Indeed it is

tombettany:
Small comment but otherwise :+1:

convolu:
Corrected as per your suggestion @tombettany, thanks

pazdera:
:shipit:

convolu:
Thank you for reviewing.

PR #178, opened 2016-01-29
Python kano signal

Ealdwulf:
@skarbat Here is the converted kano-signal. It still needs fixing as per https://github.com/KanoComputing/peldins/issues/2262 , but I think it works as far as current fucntionality goes.


skarbat:
Thanks @Ealdwulf ! Tested on the RPi and seems to work fine.
Feel free to merge.

PR #181, opened 2016-02-02
Added capability for kano-signal to switch focus

skarbat:
* Once the signal has been sent, the user focus
   is directed to the app. Otherwise a confirmation message
   box might become obscured and the user confused.
 * Removed text prints to the console via a "verbose" flag, false by default.

@Ealdwulf @tombettany

### DIFF
--- a/bin/kano-signal
+++ b/bin/kano-signal
@@ -13,7 +13,7 @@
 #
 # It is also invoked from Kano World when launching a Share.
 # The chromium hook connected to kano-profile.apps module,
-# calls this script with "launch-share <xml_filename>",
+# calls this script with "launch-share <share_filename>",
 # which will open a share on the currently running app by sending a signal.
 #
 # Exit code 0 means the signal could be delivered, any other value means error,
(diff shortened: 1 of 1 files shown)


bin/kano-signal
```diff
@@ -43,6 +46,31 @@ except:
...
+    }
+
+    if app in app_titles_dict:
+        # switch user focus back to the game app.
+        cmd = 'wmctrl -a "{}"'.format(app_titles_dict[app])
+        if verbose: print cmd
+        rc = os.system(cmd)
+        if rc != 0:
+            from kano.logging import logger
+            logger.error("error raising window {}".format(app_titles_dict[app]))
```
tombettany: This would probably be better with an `itervalues`:

```python
if app in app_titles_dict.itervalues():
    # switch user focus back to the game app.
    cmd = 'wmctrl -a "{}"'.format(app)
    if verbose: print cmd
    rc = os.system(cmd)
    if rc != 0:
        from kano.logging import logger
        logger.error("error raising window {}".format(app))
```

Also, it seems like this dict is only created to run through the apps list - would it not be better to just use a list?

bin/kano-signal
```diff
@@ -84,7 +112,7 @@ for APP in ["make-pong", "make-minecraft", "make-music", "make-light"]:
 
 if app is None:
     # No valid app is running
-    print "no valid app is running, returning error"
+    if verbose: print "no valid app is running, returning error"
```
tombettany: A bit strange to put this on one line.

bin/kano-signal
```diff
@@ -25,6 +25,9 @@ import sys
 import os
 import stat
 
+# verbose flag for debugging on the console
+verbose=False
```
tombettany: Is there any way to set this to `True` without changing this file, i.e. through a CLI parameter or system variable?

skarbat:
I have made some additions @tombettany @Ealdwulf  , could you please review?

bin/kano-signal
Ealdwulf: @tombettany I dont see how this change (which has been merged) is supposed to work. If this were `for app in`, then adding .itervalues() would change `app` to the value instead of the key of the dict. But it's not, it's `if app in`.

tombettany: @skarbat Sorry, I have mislead you on this. Ignore me and listen to @Ealdwulf.

skarbat:
I am merging this one to go forward, @Ealdwulf

PR #183, opened 2016-02-04
Utils: Remove old `.pyc` file

tombettany:
When migrating the `utils.py` file into the `utils` module, the
`utils.pyc` file was not deleted which meant that Python would ignore
the module and use the old `.pyc` file. Ensure that this is removed to
allow use of the module.

Should fix the updater problem:

```
'timeout' is an invalid keyword argument for this function
```

@convolu @pazdera @Ealdwulf to review.

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -2,7 +2,7 @@
 
 # postinst
 #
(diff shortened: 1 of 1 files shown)


convolu:
reluctantly :shipit:

PR #184, opened 2016-02-05
Implement open share signal for Make Art

skarbat:
* Added detection of kano draw app on the desktop, and sending signal to open a share
 * Fixed parameter parsing issue for other signals (save/load/share/make)
 * Added a dry-run option to debug from the command line

cc @alex5imon @tombettany


tombettany:
Looks good @skarbat

skarbat:
Thanks @tombettany, swift merge!

PR #185, opened 2016-02-09
Networking: Add libkano_networking library

tombettany:
Provide optimised networking functions implemented in C with a Python
interface.

@convolu @Ealdwulf to review.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -6,10 +6,11 @@ Standards-Version: 3.9.2
 Build-Depends: debhelper (>= 9), libraspberrypi-dev, libpng12-dev
 
 Package: kano-toolset
-Architecture: all
+Architecture: any
 Depends: ${misc:Depends}, python-gtk2, python-webkit,
          fping, python, zenity, bc, ifplugd, python-yaml, python-pyinotify,
-         resolvconf, libpng12-0, libraspberrypi-bin, procps, x11-utils, dbus, wmctrl
+         resolvconf, libpng12-0, libraspberrypi-bin, procps, x11-utils, dbus,
+         wmctrl, libc-bin
 Replaces: kano-init (<< 1.0-46)
 Breaks: kano-init (<< 1.0-46)
 Description: Collection of tools for Kanux
--- a/libs/kano-networking/includes/ifaces.h
+++ b/libs/kano-networking/includes/ifaces.h
@@ -1,0 +1,20 @@
+/**
+ * ifaces.h
+ *
+ * Copyright (C) 2016 Kano Computing Ltd.
+ * License: http://www.gnu.org/licenses/gpl-2.0.txt GNU GPL v2
+ *
+ * Methods for interracting with interfaces
+ */
+
+#ifndef __KANO_NETWORKING_IFACES__
+#define __KANO_NETWORKING_IFACES__
+
+enum ERRORS {
+    E_NO_INTERFACES = -1
+};
+
+int select_iface(char * iface_type, char **iface_name);
+int check_iface_type(char * iface_type);
+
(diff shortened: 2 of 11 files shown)


libs/kano-networking/src/ifaces.c
```diff
@@ -0,0 +1,73 @@
...
+
+    freeifaddrs(iface_addr);
+
+    return 0;
+}
+
+/**
+ * Check for interfaces of the provided type and print the interface name
+ */
+int check_iface_type(char * iface_type)
```
convolu: Are you altering the iface_type string anywhere? Could it be a `const char *` for clarity?

libs/kano-networking/src/ifaces.c
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Search through the available interfaces and return the first available
+ * which matches the type provided
+ *
+ * The caller is responsible for freeing iface_name
+ *
+ * Upon error, iface_name does not need freeing
+ */
+int select_iface(char * iface_type, char **iface_name)
```
convolu: Do you think that the `iface_type` can be a `const char *` ?

libs/kano-networking/includes/ifaces.h
```diff
@@ -0,0 +1,20 @@
...
+
+#ifndef __KANO_NETWORKING_IFACES__
+#define __KANO_NETWORKING_IFACES__
+
+enum ERRORS {
+    E_NO_INTERFACES = -1
+};
+
+int select_iface(char * iface_type, char **iface_name);
+int check_iface_type(char * iface_type);
```
convolu: Personally I like my typedefs, so I would typedef the enum and then have that be the return value of these two fns, but it is clear as it is anyway

tombettany:
Won't merge until release of v2.3.0

libs/kano-networking/src/ifaces.c
```diff
@@ -0,0 +1,76 @@
...
+        candidate_iface = candidate_iface->ifa_next;
+    }
+
+    if (!match) {
+        freeifaddrs(iface_addr);
+        return E_NO_INTERFACES;
+    }
+
+    // Allow for termination character
+    size_t len = strlen(candidate_iface->ifa_name) + 1;
```
Ealdwulf: I think these three lines are the same as  `*iface_name = strdup(candidate_iface->ifa_name);` ?

Ealdwulf:
Apart from minor point above, looks good :+1:

debian/control
```diff
@@ -6,10 +6,11 @@ Standards-Version: 3.9.2
 Build-Depends: debhelper (>= 9), libraspberrypi-dev, libpng12-dev
 
 Package: kano-toolset
-Architecture: all
+Architecture: any
```
radujipa: Is there a difference between the two here? Most of our packages use `all`.

tombettany: `all` means that the built package can be installed on any architecture without any changes (e.g. packages only containing python scripts, shell scripts, things which do not need compilation). `any` means that the package can be installed on any architecture but needs to be rebuilt for the platform for this to be possible.

convolu: Thanks for clearing this up @tombettany I was confused as well

libs/kano-networking/src/ifaces.c
tombettany: Good spot, you are indeed right.

debian/control
radujipa: Ah right :+1: Thanks for the explanation :)

tombettany:
Merging now that we have released.

PR #186, opened 2016-02-16
Add log_exeption method which logs an exception as caught.

Ealdwulf:
This PR adds a function to kano.logging for logging a handled, but unexpected exception, so we can write code like this:
```
try:
      # risky call
 except Exception as e:
      log_exception(e, "Some informative msg") 
      # do something to clean up
```
Note that it currently assumes that all such exceptions are worthy of logger.error. 
@radujipa @tombettany


Ealdwulf:
Actually, ignore this, I've had a better idea.

PR #187, opened 2016-02-16
Add an easy way of logging a handled exception.

Ealdwulf:
This PR adds an 'exception' keyword arg to the loggiing methods. This can be passed an exception object and will add 'exception' and 'traceback' fields to the log json. This allows wone to write:
```
try:
   # risky code   
except Exception as e:
   logger.error('informative message',exception = e)
   # handle the exception
``` 
@radujipa @tombettany

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,9 @@
+kano-toolset (2.3.0-6) unstable; urgency=low
+
+  * Added simple way of logging an exception.
+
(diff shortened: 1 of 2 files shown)


radujipa:
This was long overdue, nice one @Ealdwulf :+1:

Ealdwulf:
Thanks @radujipa, merging...

PR #188, opened 2016-02-16
Added RPi performance scoring utils

radujipa:
Related to https://github.com/KanoComputing/Kerberos/issues/4

### DIFF
--- a/kano/utils/__init__.py
+++ b/kano/utils/__init__.py
@@ -10,30 +10,45 @@ __author__ = 'Kano Computing Ltd.'
 __email__ = 'dev@kano.me'
 
 
+from kano.utils.audio import \
+    play_sound, percent_to_millibel, get_volume
+
+from kano.utils.disk import \
+    get_free_space, get_partition_info
+
 from kano.utils.file_operations import \
     read_file_contents, write_file_contents, read_file_contents_as_lines, \
     delete_dir, delete_file, ensure_dir, list_dir, chown_path, read_json, \
     write_json, open_locked, sed
-from kano.utils.user import \
-    get_user_getpass, get_user, get_user_unsudoed, get_home, \
-    get_home_by_username, get_all_home_folders, enforce_root
-from kano.utils.audio import \
-    play_sound, percent_to_millibel, get_volume
-from kano.utils.hardware import \
-    detect_kano_keyboard, is_model_a, is_model_b, is_model_b_plus, \
-    is_model_2_b, get_rpi_model, is_monitor, get_mac_address, get_cpu_id
-from kano.utils.processes import \
-    kill_child_processes, is_running, get_program_name, pkill
+
 from kano.utils.gui import \
     is_gui, zenity_show_progress
+
+from kano.utils.hardware import \
+    detect_kano_keyboard, is_model_a, is_model_b, is_model_b_plus, is_model_zero, \
+    is_model_2_b, get_rpi_model, is_monitor, get_mac_address, get_cpu_id, \
+    has_min_performance
+
+from kano.utils.hardware import \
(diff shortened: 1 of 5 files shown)


radujipa:
@skarbat I've added a few more tests as well for the additions to `get_rpi_model`. Not entirely sure how I'd test `has_min_performance` however. Any suggestions?

kano/utils/hardware.py
```diff
@@ -10,8 +10,28 @@
 from kano.utils.file_operations import read_file_contents_as_lines
 
 
-def detect_kano_keyboard():
+# "performance" scores for RPi boards
+RPI_A_SCORE = 1000
+RPI_A_PLUS_SCORE = 1000
+RPI_B_SCORE = 2000
+RPI_B_PLUS_SCORE = 2000
+RPI_ZERO_SCORE = 3000
```
Ealdwulf: Why is the zero greater than the B/B+?
Why is the CM greater?

Ealdwulf:
Apart from commen above, looks good to me :+1:

kano/utils/hardware.py
radujipa: Good question! From https://en.wikipedia.org/wiki/Raspberry_Pi it mentioned that the CPU on the Zero was slightly overclocked by default (1Ghz vs 900Mhz on the B/B+). Also for the Compute module, I considered io performance since the same wikipedia page mentions it has 4 GB eMMC flash memory chip. Would you suggest changing the values?

Ealdwulf: Hmm. Actually the CM might be better than the pi2 under some circumstances :-)
I guess it's okay for now since all this is internal to this module and we are only using RPI_2_B_SCORE elsewhere anyway.

kano/utils/__init__.py
```diff
@@ -10,30 +10,45 @@
...
+    kill_child_processes, is_running, get_program_name, pkill
+
+from kano.utils.shell import \
+    run_cmd, run_cmd_log, run_bg, run_term_on_error, run_print_output_error, \
+    restore_signals
+
 from kano.utils.system import \
     is_jessie, is_systemd
+
+from kano.utils.user import \
```
tombettany: Is this a result of your linter now wanting them in alphabetical order?

kano/utils/hardware.py
```diff
@@ -112,3 +138,26 @@ def get_mac_address():
...
+    rv = True
+    model = get_rpi_model()
+
+    if model in PERFORMANCE_SCORES:
+        model_score = PERFORMANCE_SCORES[model]
+
+        if model_score < score:
+            rv = False
+
+    return rv
```
tombettany: Could this just be

```python
model = get_rpi_model()
model_score = PERFORMANCE_SCORES.get(model, score)

return model_score >= score
```

or, if you don't like `.get` with default values and to make the behaviour when the board is not detected, return early

```python
model = get_rpi_model()

if model not in PERFORMANCE_SCORES:
    return True

model_score = PERFORMANCE_SCORES[model]

return model_score >= score
```

tests/test_rpi_models.py
```diff
@@ -25,7 +25,7 @@ def test_model_A(self):
         self.assertIn(getrpi('0008'), modela)
         self.assertIn(getrpi('0009'), modela)
 
-    def test_model_A_plus(self):
+    def test_module_a_plus(self):
```
tombettany: Did you mean to do this? If so, sureley `test_model_B` and `test_model_B_plus` should be changed too.

kano/utils/__init__.py
radujipa: Yup. Also my OCD is now at peace.

tests/test_rpi_models.py
radujipa: I just removed a duplicate of the same test. However, I didn't want to do more changes there.

tombettany:
:+1: if we are happy to use the supplied scores for now.

PR #190, opened 2016-02-17
DBus: Add helper functions for DBus interaction

tombettany:
Add functions for
 * Getting properties
 * Setting properties
 * Translating from DBus boolean types to python booleans

cc @convolu @radujipa


tombettany:
As there are no objections, merging in.

PR #191, opened 2016-02-17
Decorators: Add `retry` and `queue_cb` decorators

tombettany:
Add two decorators:

 * `retry`:
       Allows a function to be repeatedly called until either the
       function returns True or the number of retries is exceeded.
 * `queue_cb`:
       Causes a callback to be called upon the completion of the
       decorated function. The callback can be called with supplied
       arguments and can optionally be queued with `GObject.idle_add`.

cc @pazdera @convolu


convolu:
:shipit:

PR #192, opened 2016-02-17
Gtk: Add blur mixin

tombettany:
Add widget which supplies the loading blur effect.

cc @pazdera @convolu


tombettany:
As there are no objections, merging.

PR #193, opened 2016-02-18
Hardware Utils: Add more board properties

tombettany:
Supply, in addition to the board performance, the human readable board
name and the profile to use for clocking.

@radujipa An extension of what you did to make some properties a little easier to access.

cc @convolu

### DIFF
--- a/kano/utils/hardware.py
+++ b/kano/utils/hardware.py
@@ -9,6 +9,15 @@
 from kano.utils.shell import run_cmd
 from kano.utils.file_operations import read_file_contents_as_lines
 
+RPI_A_KEY = 'RPI/A'
+RPI_A_PLUS_KEY = 'RPI/A+'
+RPI_B_BETA_KEY = 'RPI/B (Beta)'
+RPI_B_KEY = 'RPI/B'
+RPI_B_PLUS_KEY = 'RPI/B+'
+RPI_ZERO_KEY = 'RPI/Zero'
+RPI_COMPUTE_KEY = 'RPI/Compute'
+RPI_2_B_KEY = 'RPI/2/B'
+
 
 # "performance" scores for RPi boards
 RPI_A_SCORE = 1000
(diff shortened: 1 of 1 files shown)


kano/utils/hardware.py
```diff
@@ -151,13 +226,8 @@ def has_min_performance(score):
...
-    if model in PERFORMANCE_SCORES:
-        model_score = PERFORMANCE_SCORES[model]
 
-        if model_score < score:
-            rv = False
+    model = get_rpi_model()
+    model_score = get_board_property(model, 'performance')
 
-    return rv
+    return model_score and model_score >= score
```
radujipa: Make sure you preserve the previous behaviour here please. If the above does not have a `model_score` it returns `False` instead of `True`. The assumption made there was that new boards are going up in performance instead of down.

kano/utils/hardware.py
```diff
@@ -19,18 +29,83 @@
...
+RPI_2_CPU_PROFILE = 'rpi_2'
+
+
+'''
+Lookup table with keys as given by get_rpi_model() containing:
+    * Human readable 'name' of the board
+    * 'cpu_profile' which details the settings to use
+    * 'performance' scores
+'''
+BOARD_PROPERTIES = {
```
radujipa: As this is getting fairly complicated with this new approach, I'm thinking if it is possible to have the hardware Properties and Settings all in one place. The less maintenance the better.

kano/utils/hardware.py
tombettany: Ah, gotcha - I'll make it so.

kano/utils/hardware.py
```diff
@@ -19,18 +28,80 @@
...
+def get_board_property(board_key, prop):
+    board = BOARD_PROPERTIES.get(board_key)
+
+    if not board:
+        return False
+
+    board_prop = board.get(prop)
+
+    if not board_prop:
+        return False
```
Ealdwulf: Is there a reason to return False rather than None here?

Ealdwulf:
Apart from very minor query above, looks good.

kano/utils/hardware.py
tombettany: @radujipa I absolutely agree but this looks like it could be a large task. We'll keep it like this for now with the goal to merge them at some point.

tombettany:
Merging in its current state for testing.

PR #194, opened 2016-02-19
Moved the updater into the button handler

radujipa:
Related to https://github.com/KanoComputing/peldins/issues/2067

### DIFF
--- a/bin/kano-shutdown
+++ b/bin/kano-shutdown
@@ -10,25 +10,26 @@
 # /sbin/poweroff, reboot, and loginctl.
 #
 
-# The steps needed before showing the dialog are a bit time consuming
-# so we raise the hourglass as soon as possible
-from kdesk.hourglass import hourglass_start, hourglass_end
-hourglass_start('kano-shutdown')
 
 from gi.repository import Gtk, Gdk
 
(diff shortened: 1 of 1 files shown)


Ealdwulf:
:+1:

bin/kano-shutdown
```diff
@@ -10,25 +10,26 @@
...
 from kano.gtk3.apply_styles import apply_styling_to_screen
-from kano.utils import run_cmd, run_bg
 
 from kano.paths import common_css_dir
+from kano.utils import run_cmd, run_bg
+
+# The steps needed before showing the dialog are a bit time consuming
+# so we raise the hourglass as soon as possible
+from kdesk.hourglass import hourglass_start, hourglass_end
+hourglass_start('kano-shutdown')
```
tombettany: Was this not first so that the hourglass showed while the potentially lengthy imports were loaded?

radujipa: I tested that and it doesn't make a difference, there are no lengthy imports here.

tombettany: Fair enough

tombettany:
:shipit:

radujipa:
Thanks for the review guys :+1:

PR #196, opened 2016-02-23
Add arch entry to board type.

Ealdwulf:
Add an arch field to the board definitions for use in determining when chromium will run. @tombettany


tombettany:
:shipit:

PR #197, opened 2016-02-25
Fix when Quality = ''

darryllee:
Some APs are returning Quality values of single integers instead of x/100. I'm not sure why, but getCellQuality doesn't seem to recognize a single integer, and so Quality ends up being set to '', per the default value set in line 157.

Unfortunately, when sortNetworks then tries to figure out a sorting factor, it assumes there will be a slash, and we get the error: Unhandled exception 'need more than 1 value to unpack' at /usr/lib/python2.7/dist-packages/kano/network.py line 225

I couldn't figure out why getCellQuality isn't able to deal with a single integer Quality, but I was able to workaround these weird APs (luckily neither of them is mine) by setting factor to 0 if Quality = ''.

### DIFF
--- a/kano/network.py
+++ b/kano/network.py
@@ -221,8 +221,11 @@ class IWList():
         '''
 
         def sortNetworks(adict):
-            x, z = adict['quality'].split('/')
-            factor = int(x) / float(z)
+            if adict['quality'] != '':
+                x, z = adict['quality'].split('/')
+                factor = int(x) / float(z)
+            else:
+                factor = 0
             return factor
 
         def add_wnet(wlist, new_wnet):


tombettany:
Ah, great job @darryllee! Thanks for looking into this problem; it is always very difficult to test on every AP.

It looks like the `getCellQuality()` function only gets run [if the line starts with `Quality=`](https://github.com/KanoComputing/kano-toolset/blob/cac1a3f040a852155c527bed94972560fa244e0b/kano/network.py#L180-L183) and the [`getCellQuality()`](https://github.com/KanoComputing/kano-toolset/blob/cac1a3f040a852155c527bed94972560fa244e0b/kano/network.py#L144-L146) function is looking specifically for the `=`.

If [your comment](https://github.com/KanoComputing/kano-toolset/issues/198#issuecomment-188643458) about the output is correct, those APs which only show a single number have the line being

```
Quality:40  Signal level:0  Noise level:0
```

whereas the other APs have the line

```
Quality=72/100  Signal level=65/100
```

i.e. they have `:` instead of `=`.

While this PR is good in that it prevents crashing in the case of certain APs, it isn't ideal as it doesn't allow the ranking to work. Is it possible to make some changes to [`network.py`](https://github.com/KanoComputing/kano-toolset/blob/jessie/kano/network.py) to resolve the root of this problem?

Also, we are currently using the [`jessie` branch](https://github.com/KanoComputing/kano-toolset/tree/jessie) for our development so ideally this would be rebased on that and the PR submitted to that branch.

darryllee:
Oh dur. Right a : instead of an =. It's right there. Sure let me take a
stab at fixing it properly.

We'll still need to adjust for a non fractional Quality but I actually
wrote some code for that before I realized Quality wasn't getting set at
all in those cases.

Thanks Tom!

On Thursday, February 25, 2016, Tom <notifications@github.com> wrote:

> Ah, great job @darryllee <https://github.com/darryllee>! Thanks for
> looking into this problem; it is always very difficult to test on every AP.
>
> It looks like the getCellQuality() function only gets run if the line
> starts with Quality=
> <https://github.com/KanoComputing/kano-toolset/blob/cac1a3f040a852155c527bed94972560fa244e0b/kano/network.py#L180-L183>
> and the getCellQuality()
> <https://github.com/KanoComputing/kano-toolset/blob/cac1a3f040a852155c527bed94972560fa244e0b/kano/network.py#L144-L146>
> function is looking specifically for the =.
>
> If your comment
> <https://github.com/KanoComputing/kano-toolset/issues/198#issuecomment-188643458>
> about the output is correct, those APs which only show a single number have
> the line being
>
> Quality:40  Signal level:0  Noise level:0
>
> whereas the other APs have the line
>
> Quality=72/100  Signal level=65/100
>
> i.e. they have : instead of =.
>
> While this PR is good in that it prevents crashing in the case of certain
> APs, it isn't ideal as it doesn't allow the ranking to work. Is it possible
> to make some changes to network.py
> <https://github.com/KanoComputing/kano-toolset/blob/jessie/kano/network.py>
> to resolve the root of this problem?
>
> Also, we are currently using the jessie branch
> <https://github.com/KanoComputing/kano-toolset/tree/jessie> for our
> development so ideally this would be rebased on that and the PR submitted
> to that branch.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/KanoComputing/kano-toolset/pull/197#issuecomment-188807211>
> .
>


-- 

--Darryl Lee <lee@darryl.com>

Issue #198, opened 2016-02-25
Weird APs have single integer Quality instead of x/100, which network.py barfs on


darryllee:
So in the attached iwlist.out file, you'll see a few of my neighbor's APs (grr, crappy AT&T routers) have a Quality of 12 or 19, instead of the expected 72/100 or 48/100.

Apparently network.py never expected such lame access points, but instead of just setting Quality to 12, it falls through, and it gets set to your default of '', which then breaks the sortNetworks function, which really wants to find a slash and do some division.

I wrote a workaround: https://github.com/KanoComputing/kano-toolset/pull/197

But I think it'd be better to figure out why it's not able to get the single integer Quality from these lame APs (I'm not seeing any other places where it's looking for a / or a 100), but maybe this will help somebody who's also running into this problem.

Also, I'm using a super-cheap Sabrent dongle, in case that matters. But the iwlist scan runs fine on it, so I think it's these weird APs that is throwing things off.

darryllee:
File attach in Midori not so good. Here's a copy-paste of the iwlist output I'm referring to above:

wlan0     Scan completed :
          Cell 01 - Address: 80:37:73:FA:17:CF
                    ESSID:"ATTKcVsCFA_EXT"
                    Protocol:IEEE 802.11bgn
                    Mode:Master
                    Frequency:2.437 GHz (Channel 6)
                    Encryption key:on
                    Bit Rates:144 Mb/s
                    Extra:wpa_ie=dd1a0050f20101000050f20202000050f2020050f20401000050f202
                    IE: WPA Version 1
                        Group Cipher : TKIP
                        Pairwise Ciphers (2) : TKIP CCMP
                        Authentication Suites (1) : PSK
                    Extra:rsn_ie=30180100000fac020200000fac02000fac040100000fac020000
                    IE: IEEE 802.11i/WPA2 Version 1
                        Group Cipher : TKIP
                        Pairwise Ciphers (2) : TKIP CCMP
                        Authentication Suites (1) : PSK
                    IE: Unknown: DD8D0050F204104A0001101044000102103B000103104700102880288028801880A880803773FA17CF102100044E54475210230008574E3330303052501024000256321042000C3830333737336661313763661054000800060050F204000110110017574E333030305250763228576972656C6573732041502910080002210C103C0001011049000600372A000120
                    Quality:40  Signal level:0  Noise level:0
          Cell 02 - Address: C0:FF:D4:8C:00:00
[...]

PR #199, opened 2016-02-26
Fix for Quality/Signal level output variations

darryllee:
As mentioned in https://github.com/KanoComputing/kano-toolset/issues/198, the format of Quality and Signal level changes when iwlist cannot get a range for some Access Points (or perhaps it's related to the wireless driver). At any rate, I've observed a 2Wire/AT&T U-verse router showing:

`Quality:40  Signal level:0  Noise level:0`

The original parser only accounts for Quality/Signal level in this format:

`Quality=72/100  Signal level=65/100`

This change accommodates the possibility of for a colon in addition to an equals sign, and also can handle an integer value for Quality as opposed to a fraction.

### DIFF
--- a/kano/network.py
+++ b/kano/network.py
@@ -131,7 +131,7 @@ class IWList():
                 return s.strip().split(":")[1]
 
             def getCellSignal(s):
-                s = s.split("Signal level=")[1]
+                s = re.split('Signal level(=|:)', s)[2]
                 return s.strip().split(" ")[0]
 
             def getCellNoise(s):
@@ -142,7 +142,7 @@ class IWList():
                     return 0
 
             def getCellQuality(s):
-                s = s.split("=")[1]
+                s = re.split('(=|:)', s)[2]
                 return s.strip().split(" ")[0]
 
             def getCellMAC(s):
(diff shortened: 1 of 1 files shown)


kano/network.py
```diff
@@ -233,8 +233,11 @@ def getList(self, unsecure=False, first=False, debug=False):
         '''
 
         def sortNetworks(adict):
-            x, z = adict['quality'].split('/')
-            factor = int(x) / float(z)
+            if '/' in adict['quality']:
+                x, z = adict['quality'].split('/')
+                factor = int(x) / float(z)
+            else:
+                factor = int(adict['quality']) / 100
```
tombettany: I think that you need to either `float(adict['quality'])` or use `100.`, else the factor will almost certainly always be an integer and hence 0.

tombettany:
Other than the small comment above - it looks good. After the above is fixed, I will test it.

kano/network.py
darryllee: Ah, I'm not familiar with the intricacies of float v integer operations in Python (old Perl hacker). Adding a decimal point is an elegant fix, but since I'm a noob I'll use the float() to make it more obvious. Thanks!

darryllee:
So uh, hey, my son wanted to mess around with Kano after some time away, and I didn't know where the right card was, so I downloaded the latest Beta, 3.15.0. Unfortunately we ran into the same bug that I filed 2 years ago. And my changes fixed the problem (again). Can somebody *please* merge this? I'd be surprised that this hasn't stopped others from being able to get on wireless. :-{

PR #200, opened 2016-02-29
Boards: Add Raspberry Pi 3

tombettany:
Add identifying features for the Raspberry Pi 3.

cc @radujipa @alex5imon @Ealdwulf

### DIFF
--- a/kano/utils/hardware.py
+++ b/kano/utils/hardware.py
@@ -17,6 +17,7 @@ RPI_B_PLUS_KEY = 'RPI/B+'
 RPI_ZERO_KEY = 'RPI/Zero'
 RPI_COMPUTE_KEY = 'RPI/Compute'
 RPI_2_B_KEY = 'RPI/2/B'
+RPI_3_KEY = 'RPI/3'
 
 
 # "performance" scores for RPi boards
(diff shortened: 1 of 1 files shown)


kano/utils/hardware.py
```diff
@@ -177,6 +186,8 @@ def get_rpi_model(revision=None):
             model_name = RPI_2_B_KEY
         elif int(revision, 16) & 0x00FFFFFF == 0x00900092:
             model_name = RPI_ZERO_KEY
+        elif revision.strip() == 'a02082':
```
radujipa: You may want to be careful here because of the Overclocking flag. Try to use the same approach as for the other models.

tombettany: Thanks, I thought that it would be a shortcut but I guess not...

I'm fixing now

kano/utils/hardware.py
```diff
@@ -90,6 +93,12 @@
         'cpu_profile': RPI_2_CPU_PROFILE,
         'performance': RPI_2_B_SCORE,
         'arch': "armv7"
+    },
+    RPI_3_KEY: {
+        'name': 'Raspberry Pi 3',
+        'cpu_profile': RPI_3_CPU_PROFILE,
+        'performance': RPI_3_SCORE,
+        'arch': "armv7"
```
Ealdwulf: Actually it's v8. Not that this will make a difference, as I carefully made chromium conditional on !=v6

radujipa:
:shipit:

PR #201, opened 2016-03-08
Add kano-dispmanx-launch

Ealdwulf:
Add 'kano-dispmanx-launch' command for launching dispmanx apps which need to launch other dispmanx apps with no overlap.
This is to address https://github.com/KanoComputing/kano-overworld/issues/13

It expects 'TO_LAUNCH:' commands in stdout of the apps it launches. it copies stdout to its own stdout, but currently it does so only after the app has finished, reducing the utility of stdout for debugging. This could be fixed or we could use stderr.

@alex5imon @radujipa

### DIFF
--- a/bin/kano-dispmanx-launch
+++ b/bin/kano-dispmanx-launch
@@ -11,6 +11,8 @@
 # These are queued to run next.
 #
 # Usage: kano-dispmanx-launch cmd
+# Example: kano-dispmanx-launch "/usr/bin/love-0.10 /usr/bin/kanoOverworld.love"
+
 import subprocess
 import sys
 
--- a/bin/kano-dispmanx-launch
+++ b/bin/kano-dispmanx-launch
@@ -1,0 +1,29 @@
+#!/usr/bin/env python
+
+# kano-dispmanx-launch
+#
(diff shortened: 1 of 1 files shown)


Ealdwulf:
Note that currently this doesn't cope with apps returning an error code (subprocess.check_output raises an exception in this case, which we don't catch).

bin/kano-dispmanx-launch
```diff
@@ -0,0 +1,29 @@
...
+#
+# Copyright (C) 2016 Kano Computing Ltd.
+# License: http://www.gnu.org/licenses/gpl-2.0.txt GNU General Public License v2
+#
+# Allow dispmanx apps to launch each other without overlapping.
+# We run the first command, then scan its output for lines like
+# TO_LAUNCH:cmd
+# These are queued to run next.
+#
+# Usage: kano-dispmanx-launch cmd
```
alex5imon: could you add here an example of usage? Just for documentation purposes

Ealdwulf:
@alex5imon : Okay, I've added an example.

radujipa:
:shipit:

PR #202, opened 2016-03-10
Fix kano-splash to wait for child.

Ealdwulf:
This PR causes kano-splash to wait for its child process rather than returning after the splash is finished.
This should prevent the problem with splashes in kano-overworld & kano-dashboard
@alex5imon @skarbat @tombettany

### DIFF
--- a/kano-splash/kano-splash.c
+++ b/kano-splash/kano-splash.c
@@ -32,6 +32,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <sys/types.h>
+#include <sys/wait.h>
 #include <unistd.h>
 #include <signal.h>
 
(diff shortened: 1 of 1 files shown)


radujipa:
Awwww yiisss :shipit:

skarbat:
That's wonderful @Ealdwulf ! Will apps like the one below function as expected, in now synchronous operation?

 * https://github.com/KanoComputing/kano-profile/blob/jessie/bin/kano-profile-gui#L92-L93

Ealdwulf:
@skarbat Yes, kano-profile-gui, when run from the command line, should now only return to the prompt when the app is finished. kano-stop-splash should only close the splash display, not cause the app to return to the command line.
I am assuming that nothing is relying on kano-splash to background apps, because before this PR it doesn't return anyway until the splash display is finished.

skarbat:
Thanks for clarifying @Ealdwulf . Looks great to me then!

Ealdwulf:
Thanks @skarbat, will merge for testing then...

PR #205, opened 2016-04-05
Splash fix not to keep dispmanx access open.

Ealdwulf:
This should stop kano-splash from keeping dispmanx open over the lifetime of the command it launches. 
@tombettany @skarbat  please review (I've split some refactoring into separate commits as they obscure the real change, which is in the final commit).


Ealdwulf:
Merging for workshop, but feel free to review...

PR #207, opened 2016-04-20
Lazy imports

Ealdwulf:
This might be somewhat controversial.

I want to reduce the cost of running `kano-settings-cli get overscan`. This PR reduces the cost of importing things from kano.utils and kano.logging. It gets it in total from 0.7s to 0.2s (along with https://github.com/KanoComputing/kano-settings/issues/372). However it does make it rather more complicated. Opinions @alex5imon @radujipa @tombettany @skarbat ?

### DIFF
--- a/kano/_logging.py
+++ b/kano/_logging.py
@@ -1,0 +1,356 @@
+# kano.logging
+#
+# Copyright (C) 2014 Kano Computing Ltd.
+# License: http://www.gnu.org/licenses/gpl-2.0.txt GNU General Public License v2
+#
+
+'''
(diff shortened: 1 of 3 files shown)


alex5imon:
Very interesting approach @Ealdwulf ...

Apart from readability, are there any other drawbacks?

Ealdwulf:
I don't think there are any other drawbacks @alex5imon .

alex5imon:
Then you have my 👍

Ealdwulf:
Thanks @alex5imon , merging...

PR #208, opened 2016-04-26
Fixed issue with Keyboard layout not always being changed

skarbat:
* The json file used to collect the keyboard settings
   is accessed concurrently from more than one thread,
   therefore it could eventually fail blindly without user notice.

 * Fixed by giving priority to the json writer, and reading the same
   json file in a small and fast retry loop.


Ealdwulf:
I am not really in favour of this kind of retrying process. It makes the error N times less likely but also  many times harder to debug when it does bite us. Especially as in this case we are modifying not the program which is having a problem, but a generic function which is probably used in many places. 

If we must do this, I'd prefer that the retry loop is in kano-settings where the problem is occurring. But is this bug serious enough that we must fix it in this way for the release, rather than leaving it until we have time to find a better solution? 
@skarbat

skarbat:
Totally agree @Ealdwulf , problem is time constraints to get this fixed for next release. The json writer is triggered from kano-settings itself on a separate Gtk thread, and eventually the file is zero bytes in lenght. We could move this fix into kano-settings instead...

skarbat:
@Ealdwulf I am discarding this PR in favor of the linked one above.

PR #209, opened 2016-05-05
Parson: Add Parson library

tombettany:
Make a shared library out of the Parson JSON library by Krzysztof Gabis
(kgabis@gmail.com) to facilitate reuse.

We use this in increasingly more places so it makes sense to consolidate it.

cc @skarbat @Ealdwulf


tombettany:
Merging for testing

PR #213, opened 2016-05-13
Add __init__.py so the overnight tests will run these tests.

Ealdwulf:
@skarbat @tombettany 
kano-toolset has some tests but these are not running overnoght because there is no `__init__.py`. 
Is there a reason why these tests should not run?

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,0 +1,9 @@
+#
+# __init__.py
+#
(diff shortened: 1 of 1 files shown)


skarbat:
Had not seen them before, but they should run.
Nice trick to avoid pulling /proc in the RPI model tests!  👍 @Ealdwulf

Ealdwulf:
Okay, I will merge it, and we will see if they pass :-)

PR #214, opened 2016-05-13
Set user language of WebKit instance

japonophile:
This small modification is needed for localization, to allow applications to be displayed according to the user-defined language in Kano (see make-art Japanese translation).


japonophile:
Created PR #216 based off of jessie branch.  Closing this one.

PR #215, opened 2016-05-17
Addition of new method KanoDialog.close()

skarbat:
* For apps that need to use several dialogs without stacking them up.


tombettany:
Looks good @skarbat

PR #216, opened 2016-05-17
Set user language of WebKit instance (jessie)

japonophile:
- same content as PR #214 based off of jessie branch

This change is needed for localization, to allow applications to be displayed according to the user-defined language in Kano (see make-art Japanese translation).

### DIFF
--- a/kano/webapp.py
+++ b/kano/webapp.py
@@ -299,5 +299,5 @@ class WebApp(object):
         (langcode, charset) = LANG.split('.') if '.' in LANG else (LANG, '')
         (language, country) = langcode.split('_') if '_' in langcode else (langcode, '')
 
-        return language + '-' + country
+        return language + '-' + country if country else language
 
--- a/kano/webapp.py
+++ b/kano/webapp.py
@@ -94,7 +94,7 @@ class WebApp(object):
         view.connect('close-web-view', self._close)
         view.connect('onload-event', self._onload)
 
-        language = os.getenv('LANG').replace('_', '-')
+        language = self._get_browser_lang()
         view.execute_script("window.navigator.userLanguage = '%s'" % language)
 
         if self._inspector:
@@ -288,3 +288,16 @@ class WebApp(object):
         import ast
 
         return ast.literal_eval(s) if isinstance(s, basestring) else s
+
+    @staticmethod
+    def _get_browser_lang():
+        '''
+        Get brower language (BCP47), based on OS LANG environment variable (ISO15897).
+        '''
+        LANG = os.getenv('LANG')
+
+        (langcode, charset) = LANG.split('.') if '.' in LANG else (LANG, '')
+        (language, country) = langcode.split('_') if '_' in langcode else (langcode, '')
+
+        return language + '-' + country
+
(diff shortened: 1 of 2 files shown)


Ealdwulf:
Thanks @japonophile , this is really useful. I think we could merge this but I'm leaving it open for the time being in case I get time to look up whether there is a better way to convert the 
ISO15897 language ID from $LANG to the BCP47 one used by the browser. I had a look for whether there is already code to do that but I didn't find any. I think we should at least  to remove the `.UTF8' suffix, as the browser ID doesn't have it so it might think it's part of the region code.

japonophile:
Thanks @Ealdwulf for the advice. I also did my own research but couldn't really find existing code to do this.  I found some code based on country dictionary files, but this seems a bit overkill for what we want to do.  But I agree the charset suffix would at least need to be excluded.

I have revised my change.  Could you have a look?

japonophile:
Questions / points to discuss:
- Should the function take the language ID to convert as argument, and be just a conversion function?  I took the approach of getting the language ID from OS environment inside the function instead of passing it as argument.  In the future, we may want to look at multiple environment vars.
- Should the function be extracted to a separate module? (e.g. it could be part of kano-i18n)  For now, I left it inside kano/webapp.py.

jkimbo:
Hi @japonophile, I'm picking up some of the translation work for Kano and so I'll be going through a few of your pull requests starting with this one (starting small!)

Also just wanted to say thank you for all your work on spearheading the internationalisation of the apps. It's made my job a lot easier!

kano/webapp.py
```diff
@@ -94,6 +94,9 @@ def run(self):
         view.connect('close-web-view', self._close)
         view.connect('onload-event', self._onload)
 
+        language = self._get_browser_lang()
+        view.execute_script("window.navigator.userLanguage = '%s'" % language)
```
jkimbo: I don't think this actually has any affect since according to the [documentation](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/language) for `navigator.language` it's a read only property.

The users language should be setup correctly if the user's language is set in the operating system anyway. What was the issue you were having with it?

japonophile:
@jkimbo Thanks a lot for picking it up!
Let me know if anything needs to be corrected or improved.

jkimbo:
@japonophile if you could let me know what you think of my comment: https://github.com/KanoComputing/kano-toolset/pull/216#discussion_r75139425 that would be great!

kano/webapp.py
japonophile: The issue was that the language was not set correctly.
It was set to English, although I had configured my Kano with ja_JP.UTF-8.
According to the doc, the property window.navigator.language is indeed read-only.
This is why I am using window.navigator.userLanguage instead, which I was able to successfully initialise.

I was not able to spot why it was not initialised properly.
But I agree if you find a way to ensure navigator.language is properly initialised, it would be easier as we would not need to do this.

jkimbo: Thats strange, it worked for me when my system is localised to `as_AR.UTF-8`. Can you show me the output of `locale` in the shell for you?

jkimbo:
@japonophile any update on this?

arifshanji:
[Pass Make Art the flag for i18n](https://trello.com/c/YAUSelOL/670-pass-make-art-the-flag-for-i18n)

arifshanji:
Tested. Looks good.

PR #221, opened 2016-06-17
[i18n] Make logger able to handle unicode strings when needed

japonophile:
These changes makes the logger able to handle unicode strings if they happen to reach the log.  The intent is to only send English messages to the log, but there may be exceptions.

One example is ```kano-updater```.  The way it is implemented (relying on AptOpProgress), causes localized strings to be used as phase names, and to sneak into the log.

In general, it is good practice to support unicode all the way down to avoid any issue.  For example, some user might use unicode in a file name that is then passed to the logger.

I have tested these changes on my Kano by running the ```kano-updater``` with Japanese enabled.

### DIFF
--- a/kano/_logging.py
+++ b/kano/_logging.py
@@ -168,7 +168,7 @@ class Logger:
             if self._app_name is None:
                 self.set_app_name(sys.argv[0])
 
-            lines = str(msg).strip().split("\n")
+            lines = unicode(msg).strip().split(u"\n")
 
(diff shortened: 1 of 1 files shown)


Ealdwulf:
Nice! 
I think we can merge this one @alex5imon @tombettany

Ealdwulf:
I have merged this, thanks @japonophile !

PR #226, opened 2016-09-07
Adding background updates to network hook

skarbat:
Needed by this PR: https://github.com/KanoComputing/os-dashboard/pull/249

### DIFF
--- a/udhcpc/kano.script
+++ b/udhcpc/kano.script
@@ -17,8 +17,9 @@ RESOLV_CONF="/etc/resolv.conf"
 
 . /usr/share/kano-toolset/logging.sh
 
-# Hook script to be called when network DHCP events occur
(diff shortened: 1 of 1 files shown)


skarbat:
@Ealdwulf @radujipa could you have a quick look? needed by the advised updates

Ealdwulf:
Makes sense. Are we going to remove the old call to the updater here: https://github.com/KanoComputing/kano-desktop/blob/b32d814009ee77f928c0b2eb1f01b8be743feb8a/kdesk/icon-hooks.sh#L116

skarbat:
I thought about that too @Ealdwulf , I don't think it hurts.

radujipa:
Looks good to me :shipit:

skarbat:
Thanks guys - merging now.

PR #227, opened 2016-09-07
New function set to retrieve Kano version information

skarbat:
* Allow to peek at version information chunks from the stamp file
 * Option to change the default stamp file for offline testing

@Ealdwulf @radujipa couldn't find them implemented elsewhere, think it's a good place here.

To aid on this PR: https://github.com/KanoComputing/os-dashboard/pull/249

### DIFF
--- a/kano/utils/system.py
+++ b/kano/utils/system.py
@@ -44,6 +44,8 @@ def is_systemd():
 def get_kano_version_stamp():
     '''
     Returns the raw string from the version stamp file
+    Kanux version stamps come in this form:
+       " Sat Sep  3 03:18:24 BST 2016 Kanux-Beta-v3.5.0-jessie"
     '''
     with open(kanux_stamp_file, 'r') as f:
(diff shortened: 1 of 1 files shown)


kano/utils/system.py
```diff
@@ -37,3 +39,47 @@ def is_systemd():
...
+    with open(kanux_stamp_file, 'r') as f:
+        return f.read().strip(' \n')
+
+
+def get_kano_version_date():
+    '''
+    Returns the date this version was built
+    '''
+    try:
+        return get_kano_version_stamp().split('-')[0].strip('Kanux')
```
Ealdwulf: This works but it's pretty obscure since it relies on none of the names of days of the week starting with a letter in 'Kanux' :-)

Ealdwulf:
Makes sense 👍

skarbat:
Thanks @Ealdwulf - I have added your comment and a sample stamp format for better reading.
Merging now.

PR #229, opened 2016-09-14
Prevent GUI thread from being blocked by UDHCPC hook processes

radujipa:
When connecting to WiFi, the kano.script is executed by the UDHCPC client.
In this script, we introduce various system hooks to the "internet connection
up" event. The latest one was kano-dashboard-sysupdates which launches the
Updater using the run_cmd command. Even though sysupdates is launched in the
background, processes spawned via run_cmd will block due to inheriting fds.

This fix says that the sysupdates should be run in the background and also
discard the standard output and error, on which it would block due to the
processes launched through run_cmd (since it uses Popen with piping
stdout and stderr).

@Ealdwulf @skarbat @tombettany

### DIFF
--- a/udhcpc/kano.script
+++ b/udhcpc/kano.script
@@ -32,12 +32,14 @@ function call_hook {
 
     # Collect and report RPi CPU ID
     if [ -x $network_hook ]; then
+        # TODO: investigate whether the process below would also need to pipe
+        # stdout and stderr to /dev/null if it uses run_cmd
         $network_hook $@ &
     fi
 
     # Download updates in the background
     if [ -x $updates_hook ]; then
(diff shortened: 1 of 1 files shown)


radujipa:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;ZpE7uXa0&#x2F;307-connecting-to-wifi-when-there-is-an-update-takes-almost-2-minutes"> Connecting to WIFI when there is an update, takes almost 2 minutes</a>

Ealdwulf:
:shipit:

radujipa:
Thanks for the help with this one @Ealdwulf  👍

PR #231, opened 2016-09-26
Fix for comparison between str and unicode

radujipa:
Turns out that gtk will return a str when calling get_label. This breaks the
comparison with the `_()` function which returns a unicode when it successfully
was installed. Specifically, this fix is for all Kano OS dialogs which depend
on the output returned by the dialogs `run()` function. Review with care.

@Ealdwulf @tombettany @keberlin 

For reference, this is the type of error that this fixes in `kano-settings` > `Advanced` > `Configure` (Parental lock) > change setting > hit apply > enter wrong pass :

```
kano-settings[15149] debug get_parental_enabled: True
KanoDialog: exit_dialog: ＯＫ ＯＫ <type 'str'> <type 'unicode'> <type 'unicode'>
/usr/lib/python2.7/dist-packages/kano/gtk3/kano_dialog.py:224: UnicodeWarning: Unicode equal comparison failed to convert both arguments to Unicode - interpreting them as being unequal
  if button.get_label() == _("OK"):
Error in sys.excepthook:
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/kano/_logging.py", line 365, in log_excepthook
    exc_value=exc_value)
  File "/usr/lib/python2.7/dist-packages/kano/_logging.py", line 212, in error
    self.write(msg, **kwargs)
  File "/usr/lib/python2.7/dist-packages/kano/_logging.py", line 193, in write
    self._log_file.write("{}\n".format(json.dumps(log)))
  File "/usr/lib/python2.7/json/__init__.py", line 243, in dumps
    return _default_encoder.encode(obj)
  File "/usr/lib/python2.7/json/encoder.py", line 207, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python2.7/json/encoder.py", line 270, in iterencode
    return _iterencode(o, 0)
  File "/usr/lib/python2.7/json/encoder.py", line 184, in default
    raise TypeError(repr(o) + " is not JSON serializable")
TypeError: <type 'exceptions.TypeError'> is not JSON serializable

Original exception was:
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/kano_settings/parental_config.py", line 105, in apply_changes
    if not pw_dialog.verify():
  File "/usr/lib/python2.7/dist-packages/kano_settings/parental_config.py", line 277, in verify
    if authenticate_parental_password(pw):
  File "/usr/lib/python2.7/dist-packages/kano_settings/system/advanced.py", line 129, in authenticate_parental_password
    return read_file_contents(password_file) == encrypt_password(passwd)
  File "/usr/lib/python2.7/dist-packages/kano_settings/system/advanced.py", line 125, in encrypt_password
    return hashlib.sha1(str).hexdigest()
TypeError: must be string or buffer, not int
```


Ealdwulf:
Makes sense to me 👍

radujipa:
@tombettany @Ealdwulf  I've bumped the version of this package with the intent of having `kano-settings` depend on the new version. What I'm trying to avoid with this is settings' `postinst` scripts running without the latest version of toolset which changed how the logger works.

Ealdwulf:
Good plan @radujipa . Which reminds me, could you make the logger more robust by changing [eg here ](https://github.com/KanoComputing/kano-toolset/blob/jessie/kano/_logging.py#L171) and this line to do `.encode('utf-8, 'replace')` So it won't raise an exception?

radujipa:
@Ealdwulf I've added the change 👍  also, tested on the pi under the breaking and non-breaking conditions from before and after the fixes in settings as well

radujipa:
Thanks guys 👍

PR #232, opened 2016-09-26
Keyboard detection

Ealdwulf:
This PR: adds a new function detect_kano_keyboard_type() which 
returns what kind of kano keyboard is plugged in, if any.
Modifies detect_kano_keyboard() to use it.
Note that it is untested on the spanish keyboard due to the fact that we don't have the samples in the office yet.
@skarbat @tombettany

### DIFF
--- a/kano/utils/hardware.py
+++ b/kano/utils/hardware.py
@@ -117,20 +117,23 @@ def get_board_property(board_key, prop):
     return board_prop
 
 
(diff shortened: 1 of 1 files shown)


skarbat:
Tested with sample ID and works good @Ealdwulf

Ealdwulf:
Thanks @skarbat , I'll merge for testing.

PR #234, opened 2016-10-05
Fixes for unit tests:

Ealdwulf:
This PR:
 Adds an API call to set log dir for test purposes, uses it in the test.
 Changes to actual returned code for compute module.
@skarbat @tombettany

### DIFF
--- a/kano/_logging.py
+++ b/kano/_logging.py
@@ -55,6 +55,12 @@ LEVELS = {
 }
 
 
+def _test_override_logs_dir(logs_dir):
+    # For unit test purposes only
+    global USER_LOGS_DIR
+    USER_LOGS_DIR = logs_dir
+
+
 def normalise_level(level):
     '''
     Normalise the input string, i.e.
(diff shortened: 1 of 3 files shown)


skarbat:
Is it not possible to override the environment variable from the test modules instead?
Otherwise looks good @Ealdwulf

Ealdwulf:
What environment variable, @skarbat ?

skarbat:
Sorry @Ealdwulf , I mean to say, here:

 * https://github.com/KanoComputing/kano-toolset/blob/67b1df7b6ac5bfa63fd3a44fa6246df75bc9f7f0/kano/_logging.py#L45

if it could do something like `os.getenv('USER_LOGS_DIR')` then you wouldn't need to insert a test function in the core logging library, is it?

Ealdwulf:
I just had a go at using an environment variable instead, but it would be a more invasive change, because at the moment USER_LOGS_DIR is a global, but the tests assume you can change it after initialisation. We've have to look at the environment variable everywhere USER_LOGS_DIR is used, not just at init time. Or move it, and other functions,  into the Logger Class. I'd rather not do that right now.

skarbat:
Oh... never mind @Ealdwulf - let's go 👍

Ealdwulf:
Thanks for the thinking anyway @skarbat :-)

PR #238, opened 2016-11-16
Fix logging of exception objects

Ealdwulf:
This PR fixes logging of exceptions. These are objects and must be converted to strings before 
encoding as utf-8.
@radujipa @tombettany


Ealdwulf:
If no-one has any objections, I'lll merge this.

PR #240, opened 2017-01-26
Build and export messges.pot to a separate package

Ealdwulf:
@tombettany


debian/control
```diff
@@ -103,3 +103,8 @@ Depends: ${misc:Depends},
          libparson (= ${source:Version})
 Description: Dev version of JSON library
  Dev version of JSON library by Krzysztof Gabis
+
+Package: kano-toolset-i18n-orig
+Architecture: all
+Description: Data for working on translations of kano-toolset
+Multi-Arch: foreign
```
tombettany: `Multi-Arch` here too if you want to remove it

PR #242, opened 2017-02-01
Wireless country

skarbat:
This changeset adds the capability to set the Wireless country code.
It works automatically for the `es_AR` locale, but it can also be forced via an environment var.

Tested on the Spanish image and seems to do the job - `iw reg get` returns accurate data, but I don't have the real router environment to verify.

@Ealdwulf @tombettany


skarbat:
[Add localised wifi support](https://trello.com/c/VqCwUhw5/654-add-localised-wifi-support)

kano/network.py
```diff
@@ -384,6 +384,29 @@ def wpa_conf(essid, psk, confile, wep=False):
...
+    # TODO: expect to have more mappings, i.e. en_US.UTF-8 => US
+    # Query current wireless country with "sudo iw reg get"
+    #
+    country_code=None
+    try:
+        cc=os.getenv('KANO_WIFI_COUNTRY')
+        if not cc:
+            cc=os.getenv('LANG')
+
+        cc = cc.split('_')[0]
```
Ealdwulf: Do we really want the language part (es), not the country part (AR)? Does Argentina use the same wifi channels as Spain?

skarbat: You're probably right @Ealdwulf , I was taking the wrong code from the locale.
Fixed and updated tests.

Ealdwulf:
Hmm. Okay, we're now going to get the US set of channels when the language is set to english, instead of the international set. Im not sure whether we should do that. If there are fewer channels then some people whose wifi  works will no longer work, and if there are more channels then people who are in places other than the US will potentially be interfering with things they are not supposed to.
Can you check what the difference between the international and US channels are, @skarbat ?
The US channels will be in `/sbin/regdbdump  /lib/crda/regulatory.bin `, not sure where the international set lives.

skarbat:
Interestingly enough... with my latest changes above, I now get the following with the `AR` locale set.

```
albert@albert ~ $ sudo iw reg get
[sudo] password for albert:
country ES: DFS-ETSI
	(2400 - 2483 @ 40), (N/A, 20), (N/A)
	(5150 - 5250 @ 80), (N/A, 23), (N/A), NO-OUTDOOR
	(5250 - 5350 @ 80), (N/A, 20), (0 ms), NO-OUTDOOR, DFS
	(5470 - 5725 @ 160), (N/A, 26), (0 ms), DFS
	(57000 - 66000 @ 2160), (N/A, 40), (N/A)
albert@albert ~ $ cat /etc/kano_wpa_connect.conf
network={
	ssid="obfuscated"
	#psk="obfuscated"
	psk=12345

          scan_ssid=1
          key_mgmt=WPA-EAP WPA-PSK IEEE8021X NONE
          pairwise=CCMP TKIP
         }
         ctrl_interface=/var/run/wpa_supplicant
         country=AR
```

Researching on that @Ealdwulf

Ealdwulf:
My previous comment was incorrect, @skarbat was already not changing to the US set.

skarbat:
@Ealdwulf just to clarify, I did the change in a way that the current english images should not be set to any wireless country code:

 * https://github.com/KanoComputing/kano-toolset/blob/wireless-country/tests/test_wifi_countries.py#L72-L75

However, I am a bit confused on my results above (he is switching to ES). Could it be my wireless setup maybe? I guess we should test with real hardware...

tombettany:
@skarbat @Ealdwulf What do we think about this one (ignore the Jenkins test fail - that is to be expected).

skarbat:
I think this needs more work - Last thing I followed was that there is something else in the system that's setting the wireless country at the same time.

 * /etc/default/crda
 * cfg80211 driver

skarbat:
Tried with a kit acting as an access point on channel 13 with `hostapd`.
The changeset works well when applied to our latest development image.
Without the patch, the kit does not even see the network when scanning.

Channel 13 seems to be available on [major countries](https://en.wikipedia.org/wiki/List_of_WLAN_channels#Interference_concerns) except USA: .

```
wlan0     13 channels in total; available frequencies :
          Channel 01 : 2.412 GHz
          Channel 02 : 2.417 GHz
          Channel 03 : 2.422 GHz
          Channel 04 : 2.427 GHz
          Channel 05 : 2.432 GHz
          Channel 06 : 2.437 GHz
          Channel 07 : 2.442 GHz
          Channel 08 : 2.447 GHz
          Channel 09 : 2.452 GHz
          Channel 10 : 2.457 GHz
          Channel 11 : 2.462 GHz
          Channel 12 : 2.467 GHz
          Channel 13 : 2.472 GHz
          Current Frequency:2.472 GHz (Channel 13)
```

This should be ready to merge for further testing @tombettany @Ealdwulf

skarbat:
Merging for testing tomorrow.

PR #243, opened 2017-02-07
Move expand

Ealdwulf:
Remove expand-rootfs from shutdown. That's a somewhat risky place to run it.
@tombettany


Ealdwulf:
For https://trello.com/c/6orRiN0D

PR #247, opened 2017-02-27
Removed kanux_stamp wrapper functions, file is for internal use

radujipa:
Related to https://github.com/KanoComputing/peldins/pull/2818#issuecomment-282685988

@skarbat @tombettany Are we ok with this?


skarbat:
👍

PR #248, opened 2017-02-27
Set wireless country dynamically

skarbat:
The first time the kit boots up with the default country settings, it won't be able to scan non standard channels (13 and above), hence you won't be able to reach and connect to your network.

This changeset fixes that, in order for `kano-wifi-gui` to display all visible networks on supported channels when presenting the scanned list.

@radujipa @tombettany


skarbat:
[Add localised wifi support](https://trello.com/c/CXPqHZ7g/57-add-localised-wifi-support)

kano/network.py
```diff
@@ -404,10 +410,25 @@ def wpa_conf(essid, psk, confile, wep=False):
...
         if cc and cc not in ('US'):
             country_code=cc.upper()
-
     except:
         pass
 
+    # Tell the wireless driver to operate on this country
+    # Allows to scan for more channels where available
+    if country_code and enable_driver:
+        print 'yeaaaah'
```
radujipa: Debug message could be removed

skarbat:
Thanks Radu! :-)

PR #249, opened 2017-03-07
Fixed typo in function name causing exceptions in QA tests

radujipa:
@skarbat I think this was causing the issue. 

Tracebacks come from the same line:
```
======================================================================
ERROR: test_force_flush_env (tests.test_logging.TestForceFlush)
Print a message for each of the levels, but force flush
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/tmp/repos/kano-toolset/tests/test_logging.py", line 67, in setUp
    klm._test_override_log_dir(self.temp_dir)
  File "/tmp/repos/kano-toolset/kano/logging.py", line 46, in __getattr__
    return getattr(lm, name)
AttributeError: 'module' object has no attribute '_test_override_log_dir'
```

Function in logging module: 
https://github.com/KanoComputing/kano-toolset/blob/master/kano/_logging.py#L58


radujipa:
[Issues with nightly QA tests](https://trello.com/c/oklwXynT/774-issues-with-nightly-qa-tests)

skarbat:
Yes, just confirmed that on the kit. 👍

PR #256, opened 2017-10-05
Changes to the network module to detect wpa_supplicant errors

skarbat:
This changeset allows the network `connect` function to detect wireless connection errors.


kano/network.py
```diff
@@ -535,6 +546,75 @@ def reload_kernel_module(device_vendor='148f', device_product='5370', module='rt
...
+    This code uses heuristics based on the supplicant source code: https://w1.fi/wpa_supplicant/
+    '''
+    rc=None
+    connected=False
+    scans = 0
+
+    def is_internet_up(monitor_file=INTERNET_UP_FILE):
+        return os.path.isfile(monitor_file)
+
+    run_cmd("wpa_supplicant -D nl80211,wext -t -d -c%s -i%s -f /var/log/kano_wpa.log -B" % (wpa_file, iface))
```
tombettany: Don't you want to use your `SUPPLICANT_CMD` template above? And also use `SUPPLICANT_LOGFILE`?

kano/network.py
```diff
@@ -535,6 +546,75 @@ def reload_kernel_module(device_vendor='148f', device_product='5370', module='rt
...
+    time.sleep(1)
+    rc=cli.poll()
+    while rc == None:
+        output = cli.stdout.readline().replace('\n', '')
+        print '>>> {} <<<'.format(output)
+
+        if output.find('CTRL-EVENT-CONNECTED') != -1:
+            print '[[[ Associated ]]]'
+            cli.stdin.write('quit\n')
+            cli.stdin.flush()
```
tombettany: Maybe factor this close WPA connection out into a function?

kano/network.py
```diff
@@ -535,6 +546,75 @@ def reload_kernel_module(device_vendor='148f', device_product='5370', module='rt
...
+        elif output.find('reason=CONN_FAILED') != -1:
+            print '[[[ AP not in range ]]]'
+            cli.stdin.write('quit\n')
+            cli.stdin.flush()
+            rc=RC_AP_NOT_IN_RANGE
+            break
+
+        if output.find('WPS-AP-AVAILABLE') != -1:
+            scans += 1
+            if scans == 5:
```
tombettany: Oh, I see, you did a timeout on number of scans. Good idea but do we also want a time-based one?

kano/network.py
```diff
@@ -535,6 +546,75 @@ def reload_kernel_module(device_vendor='148f', device_product='5370', module='rt
...
+                cli.stdin.write('quit\n')
+                cli.stdin.flush()
+                reason_code = RC_AP_NOT_IN_RANGE
+                break
+
+        rc=cli.poll()
+
+    # If we are associated, wait for DHCP lease to become available
+    if rc==0:
+        for attempt in range(0, connection_timeout):
```
tombettany: Not a big deal but use `xrange()` instead because it does the same thing but is faster.

kano/network.py
```diff
@@ -535,6 +546,75 @@ def reload_kernel_module(device_vendor='148f', device_product='5370', module='rt
...
+                print '[[[ timeout due to too much scanning, router not in range ]]]'
+                cli.stdin.write('quit\n')
+                cli.stdin.flush()
+                reason_code = RC_AP_NOT_IN_RANGE
+                break
+
+        rc=cli.poll()
+
+    # If we are associated, wait for DHCP lease to become available
+    if rc==0:
```
tombettany: Should this if not be `if rc == RC_CONNECTED`. Exactly the same here as `RC_CONNECTED = 0` but still...

kano/network.py
```diff
@@ -604,28 +687,25 @@ def connect(iface, essid, encrypt='off', seckey=None, wpa_custom_file=None, conn
...
-            return False
+        #assoc_timeout = 20  # seconds
+        #assoc_start = time.time()
+        #while (time.time() - assoc_start) < assoc_timeout:
+        #    out, _, _ = run_cmd('wpa_cli -p /var/run/wpa_supplicant/ status|grep wpa_state')
+        #    wpa_state = out.strip('\n')
+        #    if len(wpa_state) and wpa_state.split('=')[1] == 'COMPLETED':
+        #        associated = True
+        #        break
+        #    time.sleep(0.5)
```
tombettany: Given that we aren't using any of this anymore - just delete it. We have it in the git history anyway and just makes things harder to follow.

kano/network.py
```diff
@@ -535,6 +546,75 @@ def reload_kernel_module(device_vendor='148f', device_product='5370', module='rt
     return reloaded
 
 
+def connection_result(iface, wpa_file, connection_timeout, verbose=False):
```
tombettany: Maybe change the function name to reflect that it is also doing the connect, currently it sounds like it is just figuring out whether you are connected after you've already attempted to connect. Maybe something like `attempt_connection()`,  `try_connect()` or even `do_connect()`.

kano/network.py
```diff
@@ -551,7 +631,9 @@ def connect(iface, essid, encrypt='off', seckey=None, wpa_custom_file=None, conn
     and internet connection to become ready.
     '''
 
-    if os.access('/var/run/kano-connect.pid', os.R_OK):
+    wpafile=None
```
tombettany: Currently a little confusing being called the "wpa" file as it can actually be a file for both WPA and WEP connections. Maybe just call it something like `conf_file` or `wifi_conf_file`?

kano/network.py
skarbat: Totally agreed, fixed.

kano/network.py
skarbat: Ups, missed it. Fixed

kano/network.py
skarbat: Yes, done.

kano/network.py
skarbat: absolutely. fixed

kano/network.py
skarbat: Good to know that in the future! fixed.

kano/network.py
skarbat: Makes a lot of sense, done.

kano/network.py
skarbat: Fixed too

skarbat:
Updated changes as requested @tombettany .
It works well, however it needs more testing as I have been able to stall it occasionally.

skarbat:
The detected stall problem was caused by a missing DHCP event.
Tested on the PI and seems to work well now. Ready to go.

skarbat:
[(8) As a user, I want helpful error messages when my wi-fi connection fails [2]](https://trello.com/c/v4m9mTUM/1332-8-as-a-user-i-want-helpful-error-messages-when-my-wi-fi-connection-fails-2)

PR #258, opened 2017-10-13
Parental control daemon was failing to bind on port 53

skarbat:
For some reason, `systemd-run` was preventing the `kano-sentry-startup` to proceed normally and bind against port 53 to provide the DNS server.

The sentry script was confusingly returning 0 meaning success, while the port was not bound.
There is actually no need to use `systemd-run` because the script is pretty fast and the server launched in the background.

@Ealdwulf @radujipa @tombettany


skarbat:
Thank you guys!

PR #259, opened 2017-10-24
Moved kano content and tracking sync to network hooks from kano-ui-autostart

radujipa:
Rather than executing the content and tracking sync routines
before launching the Dashboard, do so when the network goes up
and without the need to check for internet.

Related to: https://github.com/KanoComputing/kano-desktop/pull/270/commits/0b7e389b614feadf2ca46efd3ed9eee712a9e50d


radujipa:
[(8) As a user, I would like my kit to boot up very quickly](https://trello.com/c/kgsbu9gZ/1301-8-as-a-user-i-would-like-my-kit-to-boot-up-very-quickly)

dhcpcd-hooks/65-kano-dhcp-hook
```diff
@@ -35,14 +34,24 @@ REBOOT | BOUND)
...
     systemd-run /usr/bin/kano-network-hook $reason
     systemd-run /usr/bin/kano-dashboard-sysupdates "bound" "$ip_addr"
+
+    # Try uploading the tracking data to our servers
+    # Should be quiet on failure
+    kano-sync --upload-tracking-data --silent &
+
+    # Sync objects from the content API in the background
+    # TODO: This should be removed once we have the daemon done
+    sudo kano-content sync &
```
radujipa: Do we believe this might cause us any problems? Running something with `sudo` from this hook?

radujipa: Update on this: seems like kano-content is able to write files with the correct permissions so it seems like it's working as expected.

skarbat: Ouch, most probably `sudo` is not needed here, the hooks run as root already

radujipa: I've double checked and you're absolutely right - `sudo` is not required and I've removed it now.

radujipa:
@skarbat I've made sure that the lines added here are run in parallel with `&`, but I was wary of running them through `systemd` given unexpected issues that has created in the past. I can test tomorrow if there are any problems, but as it is it shouldn't slow down anything.

skarbat:
Perfect @radujipa , I agree `&` should be fine.

PR #260, opened 2017-10-25
WebApp: Add Kano URI scheme comprehension

tombettany:
To interact with Kano World, a custom URI scheme was created to allow
launching of apps related to shares and the like. This scheme was not
implemented in the WebApp browser so any routing to one of these schemes
would lead to a 404 error. Prevent this happening by teaching it about
this scheme.


tombettany:
[[P-TBD] Make Art + CK - Launching existing creations from within the app goes to KW, then launching the share shows 'Unable to load page'](https://trello.com/c/Nv31g2ON/1484-p-tbd-make-art-ck-launching-existing-creations-from-within-the-app-goes-to-kw-then-launching-the-share-shows-unable-to-load-page)

tests/test_webapp.py
```diff
@@ -0,0 +1,91 @@
...
+
+        if not navigation_action.is_scheme_call():
+            assert False, 'Funciton should only be called for scheme URIs'
+
+        assert \
+            args[0] == 'systemd-run --user /usr/bin/xdg-open {}'.format(uri), \
+            'Scheme launch command is incorrect'
+
+
+    with mock.patch.object(os, 'system', side_effect=fake_os_system):
```
radujipa: You could use `monkeypatch` fixture from `pytest` instead here, but not necessary.

tests/test_webapp.py
```diff
@@ -0,0 +1,91 @@
...
+
+        def fake_func(*args, **kwargs):
+            '''
+            Stub function to ensure that the API call does the right thing
+            '''
+
+            func_called['fake_func'] = True
+
+            if not navigation_action.is_api_call():
+                assert False, 'Funciton should only be called for API URIs'
```
radujipa: *Function

tests/test_webapp.py
```diff
@@ -0,0 +1,91 @@
...
+
+            assert args == call_args, 'API function call arguments incorrect'
+
+        setattr(webapp, func, fake_func)
```
radujipa: Should this be `monkeypatch`ed instead?

tests/test_webapp.py
```diff
@@ -0,0 +1,91 @@
...
+#
+# Tests for the WebApp browser
+#
+
+
+import os
+import mock
+
+
+def test_navigation_requests(webapp, navigation_action):
```
radujipa: As a note for the future, I find it easier to test specific functions. In this case this tests `_nav_req_handler`, `_api_handler`, and `_scheme_handler`. When you scope down your tests you'll find that you'll only ever need to do one `assert` at the end as well.

tombettany:
Thanks @radujipa I've made all the changes that you've suggested and added an entry into the changelog file.

PR #262, opened 2017-11-15
Adding assets for kano-splash app load animation

skarbat:
This changeset includes the assets for the default kano-splash animation for loading apps.
Application shortcuts across the system need to refer to this folder to animate their load up.


skarbat:
[(2) As a user, I want animated loading indicators in all the apps [5]](https://trello.com/c/0zGStI14/1518-2-as-a-user-i-want-animated-loading-indicators-in-all-the-apps-5)

skarbat:
Assets packaged and affected apps modified to use the splash animation.
Tested on the PI and works well, ready for merging @radujipa

radujipa:
@skarbat Could we also replace these ones?

https://github.com/KanoComputing/terminal-quest/blob/master/bin/linux-story-gui#L1
https://github.com/KanoComputing/kano-video/blob/master/bin/kano-video-browser#L1
https://github.com/KanoComputing/make-light/blob/master/bin/make-light#L1
https://github.com/KanoComputing/kano-profile/blob/master/bin/kano-profile-customise#L1
https://github.com/KanoComputing/make-snake/blob/master/snake/snake-launch-overworld.sh#L1

Also, I would advise to not have "users" of `kano-splash` specify the path to a directory they don't install. For instance, if `make-snake` doesn't have a custom animation it installs in `/usr/share/make-snake/splash-animation/` which it would specify to `kano-splash`, then it wouldn't give any parameter to the tool. As such, `kano-splash` would use `/usr/share/kano-toolset/kano-splash/splash-animation-loader` as a default value for that parameter. So, the shebang for an app would look like `#!/usr/bin/kano-splash /usr/bin/env python` if it doesn't have a custom animation.

skarbat:
I think I see what you mean, makes sense. Given that all the apps already depend on kano-toolset, which is where kano-splash lives, I would suggest using a keyword like `loader-animation` instead of removing the parameter from the shebang. At the same time, kano-splash could still be accepting hard coded paths to animations if we ever needed to, or add more keywords for additional animations.

skarbat:
I have pushed a changeset to show what I mean, @radujipa - still need to test it though.

radujipa:
@skarbat Yup, that's exactly what I meant. Now we just need to replace `/usr/share/kano-toolset/kano-splash/splash-animation-loader` with `loader-animation` in all the app PRs.

skarbat:
Perfect! I will proceed from here.

skarbat:
All additional apps have been added, I tested them to see they work fine on the PI.
Ready for merging @radujipa

PR #263, opened 2017-11-16
Write cpuid to /boot/cpuid.txt

Ealdwulf:
This PR:
 * Changes kano-network-hook to save the cpuid to /boot/cpuid.txt as well as /etc/
   We don't put the cpuid in the filename in /boot to avoid confusing  users. Also don't write it more than once.

 * Fixes return value not to cause systemd to think the process has failed if
the cpuid was already sent.

For https://trello.com/c/9zIfnzFG


bin/kano-network-hook
```diff
@@ -29,9 +30,14 @@ def send_cpu_id():
...
-    cpuid_filename = '/etc/cpuid-{}'.format(get_cpu_id())
+    cpuid = get_cpu_id()
+    cpuid_filename = '/etc/cpuid-{}'.format(cpuid)
+    cpuid_filename_boot = '/boot/cpuid.txt'.format(cpuid)
+    utc_time = time.asctime(time.gmtime(time.time()))
+
     if os.path.isfile(cpuid_filename):
         # We have already sent the CPU ID
+        sent = True
         pass
```
skarbat: `pass` not needed anymore?

bin/kano-network-hook
```diff
@@ -40,21 +46,33 @@ def send_cpu_id():
...
         except:
             if os.path.isfile(cpuid_filename): os.unlink(cpuid_filename)
             logger.error('Error sending Kit serial number')
- 
+
+    if os.path.isfile(cpuid_filename_boot):
+        pass
+    else:
+        try:
+            with open(cpuid_filename_boot, 'a') as f:
```
skarbat: probably innocuous, but should'nt it be accessed in `w`rite instead of `a`ppend?

bin/kano-network-hook
Ealdwulf: Yes, I'll remove.

bin/kano-network-hook
Ealdwulf: True.

PR #264, opened 2017-11-22
kano-stop-splash to also terminate splash running in the background

skarbat:
Complements this PR: https://github.com/KanoComputing/os-dashboard/pull/356


bin/kano-stop-splash
```diff
@@ -15,5 +16,6 @@ if test -n "$SPLASH_PID" ; then
 	   kill -ALRM $SPLASH_PID
        fi
    fi
+else
+    pkill -f kano-start-splash
 fi
```
Ealdwulf: Is there really no way to pass SPLASH_PID in? I was hoping to use kano-start-splash to do the popups for the screen capture tool. It would be better if kano-stop-splash didn't randomly kill other popups.

skarbat: Oh, totally agreed @Ealdwulf , It's not so easy because it happens from separate parts on the OS but I'll think of a solution.

skarbat:
Alex @Ealdwulf , I have submitted a change which allows to manage multiple splash windows by name, from separate processes. Works fine on the PI. Also updated README with sample code. Below is a visual example running 2 splash windows:

kano-screenshot-thu-nov-23-16-16-35

PR #267, opened 2017-11-30
Rerouted Logout to Reboot to avoid bug in kano-boards-daemon (PiHat)

radujipa:
The issue here is that kano-boards-daemon clashes with the audio
driver (snd_bcm2835) when it's restarted due to a Logout. For the
time being, execute a Reboot when user opts to Logout which will
show the Login greeter afterwards.


radujipa:
[[P2] User switching breaks audio](https://trello.com/c/qfNlWLpL/1569-p1-broken-audio-after-greeter)

debian/changelog
```diff
@@ -4,8 +4,9 @@ kano-toolset (3.14.0-0) unstable; urgency=low
   * Added short video animations to kano-splash, see README for details
   * Add Kano URI scheme comprehension to WebApp
   * Add feature to stop kano-start-splash running in the background
+  * Rerouted Logout to Reboot to avoid bug in kano-boards-daemon (PiHat lib)
 
- -- Team Kano <dev@kano.me>  Wed, 22 Nov 2017 18:48:11 +0100
+ -- Team Kano <dev@kano.me>  Tue, 30 Nov 2017 11:26:00 +0100
```
skarbat: Thursday...

radujipa: Rats! Fixed..

PR #269, opened 2017-12-01
Implement timeout in animations, provide debug flag

skarbat:
This changeset forces the animations to stop based on a timeout.
It can be overriden with the `-t` option as well.

If fixes an eventual problem where shebang based animations could stall in eternity if they did not call `kano-stop-splash` before exiting.


skarbat:
[[P1] Onboarding: Users get stuck after KW account creation](https://trello.com/c/hLoD1ZQC/1364-p1-onboarding-users-get-stuck-after-kw-account-creation)

PR #270, opened 2017-12-04
File Utils: Make sed() command more resistant

tombettany:
If the `sed()` command exported by the utils library is cut short for
some reason (e.g. if the power is pulled) then it will have overwritten
the file that it was supposed to be `sed`ing but not populated it.
Prevent this from happening by using a temporary file and copying it to
the destination.


tombettany:
[[P1] LXDE login screen error](https://trello.com/c/fBB8Bdgu/648-p1-lxde-login-screen-error)

tombettany:
[[P1] LXDE login screen error](https://trello.com/c/fBB8Bdgu/648-p1-lxde-login-screen-error)

PR #272, opened 2017-12-19
Fixed issue with Make Snake launching windowed from Desktop

radujipa:
For some reason the issue here came from kano-launcher when it
attempts to maximise the app. Make Snake already maximises itself
when it launches, but seems to be resized by the launcher afterwards
in Desktop mode.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,9 @@
+kano-toolset (3.15.0-0) unstable; urgency=low
+
+  * Fixed issue with Make Snake launching windowed from Desktop
+
+ -- Team Kano <dev@kano.me>  Tue, 19 Dec 2017 11:09:00 +0100
+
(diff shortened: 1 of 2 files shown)


radujipa:
[As a user, I do not want Make Snake window to be windowed when launching from Desktop](https://trello.com/c/pODdEpnW/1699-as-a-user-i-do-not-want-make-snake-window-to-be-windowed-when-launching-from-desktop)

kano-read-only:
[(5) As a user who unplugs during an upgrade, I want my splash screen to show a different message indicating that the system is trying to recover](https://trello.com/c/gqsa1dWi/1765-5-as-a-user-who-unplugs-during-an-upgrade-i-want-my-splash-screen-to-show-a-different-message-indicating-that-the-system-is-tryi)

PR #278, opened 2018-01-30
Added touch() utility function

radujipa:
### Added docs to project with Jenkins CI

This is so we can start having nice docs for kano-toolset. There are still some error because of `kano.logging` and code outside of functions.

### Added touch() utility function

Using a pythonic way of creating/modifying empty files rather than
calling `touch` through a system or subprocess call.

### Disabled calls to kano.profiling

The purpose here is to avoid running profiling code in production.
This should perhaps only be enabled in a special env.


kano/utils/file_operations.py
```diff
@@ -201,6 +201,31 @@ def write_json(filepath, data, prettyprint=False, sort_keys=True):
...
+        path (str): Path to the file create/modify
+        times (tuple): See :func:`os.utime`
+
+    Returns:
+        bool: Whether the operation was successful or not
+    """
+    try:
+        with open(path, 'a'):
+            os.utime(path, times)
+    except (IOError, OSError) as error:
```
wizofe: Would you like to try and create a file in a dir as the path if it's not there? Or is it just about timestamping? 

```
(...)
except OSError:
        os.makedirs(os.path.dirname(path))
        open(path, "w").close()
        os.utime(path, times)
```

wizofe: That's right or pass `dir` as a parameter to the wrapper:

```
def touch(path, times=None, dirs=False)        
    """
    A nice description.

    Args:
        path: file path to touch
        times: tuple blah blah defaults to current time.
        dirs: is set, create folders if not exists 
    """
    
    if dirs:
        basedir = os.path.dirname(path)
        if not os.path.exists(basedir):
            os.makedirs(basedir)
            
    with open(path, 'a'):
        os.utime(path, times) 
```

radujipa: Called the function beforehand to handle that automatically 👍

radujipa:
[(1) As an engineer, I want to change the error messages for V30DB so I can learn more about what's going on [2]](https://trello.com/c/fVSfXXqG/1747-1-as-an-engineer-i-want-to-change-the-error-messages-for-v30db-so-i-can-learn-more-about-whats-going-on-2)

kano/profiling_late.py
```diff
@@ -17,14 +17,20 @@
...
-# load the configuration file
-with open(CONF_FILE, 'r') as inp_conf:
-    conf = yaml.load(inp_conf)
+conf = None
 myProfile = cProfile.Profile()
 app_name = sys.argv[0]
 point_current = ""
 
 
+def load_config():
```
tombettany: Don't we need to tell something that it should be using this or is it never used anyway? I don't see the matching part.

radujipa: As luck would have it, this was being used in production. Fixed it here and 2 more PRs incoming 👍

PR #279, opened 2018-02-01
Defining currently undescribed triggers for network hooks.

wizofe:
Implemented the `INFORM`, `RENEW`, `REBIND`, `RELEASE` hooks that caused the network to stall in case they were triggered.

After some testing by setting the above triggers to an undefined behaviour, I could successfully replicate the bug, so I am pretty certain that's the case and not through the dashboard.

A future implementation of a `daemon` (as described on the `TODO` comment in the hook script) would be the ideal solution for networking though. 

Note: Changed also the script by adding functions and commenting, as to easy future updates or changes in case of a 'daemonised' networking approach.


wizofe:
[(5) As a user, I want the UI to show me the latest status of the wi-fi connection after it times out [3]](https://trello.com/c/xsHGFGmc/1751-5-as-a-user-i-want-the-ui-to-show-me-the-latest-status-of-the-wi-fi-connection-after-it-times-out-3)

dhcpcd-hooks/65-kano-dhcp-hook
```diff
@@ -1,34 +1,32 @@
...
-    ip_addr=`ip addr show wlan0 | grep inet | awk '{print $2}' | sed -s 's/\/.*//')`
-    logger "kano-dhcp-hook: REBOOT event ipaddr=$ip_addr on iface=$interface"
-
-    # Are we already connected through a different device?
+# Check if the network is already up
+# through another interface/device
+is_connected () {
     if [ -f $monitor_file ]; then
         logger "kano-dhcp-hook: REBOOT hook discarded, already networked from a previous REBOOT"
         exit 0
```
skarbat: Should this `return` from the function instead of exiting the script?

dhcpcd-hooks/65-kano-dhcp-hook
```diff
@@ -38,24 +36,60 @@ REBOOT | BOUND)
...
-NOCARRIER | STOPPED)
-    # Are we still networked due to multihomed connection?
+# Check for a multihomed connection and keep it if there
+# otherwise disconnect with a logger notification
+is_multihomed () {
     is_internet
     if [ $? -eq 0 ]; then
-    	logger "kano-dhcp-hook: NOCARRIER/STOPPED but still connected, aborting action"
+    	logger "kano-dhcp-hook: $reason but still connected, aborting action."
     	exit 0
```
skarbat: Same as above for `return` ?

wizofe: Thanks, @skarbat. 

Indeed, I can return a value (e..g 0 for fail) from both functions and then test it inside the script to exit. Do you think this would be a more suitable generalised approach? If so I am going to change.

skarbat: Oh, true. Well, given that there is a clear log on exit I think it's all good as it is.
👍 @wizofe

dhcpcd-hooks/65-kano-dhcp-hook
```diff
@@ -38,24 +36,60 @@ REBOOT | BOUND)
...
+    kano_network_init
+    kano_daemon
+    ;;
+
+INFORM|RENEW|REBIND|RELEASE)
+    is_internet
+    is_reconnected
     ;;
 
+NOCARRIE|STOP|STOPPED)
```
radujipa: Typo here, should be `NOCARRIER`

radujipa: Should this also include `TIMEOUT`, `EXPIRE`, `DEPARTED`, `FAIL`?

dhcpcd-hooks/65-kano-dhcp-hook
```diff
@@ -38,24 +36,60 @@ REBOOT | BOUND)
...
+        logger "kano-dhcp-hook: $reason and connectivity lost."
         rm $monitor_file
     fi
+}
+
+# In case of lease RENEWal or REBOUND to a new DHCP server
+# check if connection is sucesfully active and modifly
+# the monitor_file accordingly
+is_reconnected () {
+    if [$? -eq 0]; then
```
radujipa: Should this be `[ $? -eq 0 ]` - note spaces. Will Bash be picky here?

skarbat: I seem to remember that dhcpd uses `sh` not `bash`, might be why.

dhcpcd-hooks/65-kano-dhcp-hook
```diff
@@ -38,24 +36,60 @@ REBOOT | BOUND)
...
+      logger "kano-dhcp-hook: $reason but still connected...Aborting."
+    exit 0
```
tombettany: Should this be indented to match the above line?

dhcpcd-hooks/65-kano-dhcp-hook
tombettany: Maybe the `is_internet` call should be moved into this function so that the `$?` always makes sense.

Ealdwulf: `sh` is just as picky.

dhcpcd-hooks/65-kano-dhcp-hook
Ealdwulf: Probably.

dhcpcd-hooks/65-kano-dhcp-hook
radujipa: There's a catch with these hook scripts: you cannot use `exit` otherwise all subsequent hooks will not execute.

wizofe: Yup, I am using `exit` in all cases that there is already connection present, that means hooks don't need to be triggered and the `wpa_supplicant` handles the connection as it is. Actually, there is no change there from the previous script, I only created the functions for a clearer structure.

wizofe: Yup, done.

dhcpcd-hooks/65-kano-dhcp-hook
wizofe: Both is fine for `sh` but I applied the spacing as for uniformity with the above.

dhcpcd-hooks/65-kano-dhcp-hook
wizofe: Thanks corrected.

dhcpcd-hooks/65-kano-dhcp-hook
radujipa: Oh, I see. I'm not really happy with breaking the event listener chain as I can't see what gotchas there may be. If that is who it's going to be handled, please clearly document this behaviour in the header of the file.

dhcpcd-hooks/65-kano-dhcp-hook
wizofe: On `DEPARTED` and `FAIL` the `dhcpcd` can't run anyway, as it means that there is no interface; therefore no need to be triggered (except if want to be log that through the `logger` - is that a wished feature?). 

The important and undescribed behaviour that was causing the bug, was while a `RENEW` or a `REBOUND` wasn't re-triggering the `dhcpcd` and `wpa_supplicant` as an effect.

`TIMEOUT` keeps already an old lease and in case of `EXPIRED`, it is also handled by default from the `dhclient`, so no need to re-define the default behaviour.

PR #280, opened 2018-02-05
Updated RPi board revision numbers and detection functions

radujipa:
### Updated RPi board revision numbers and detection functions

This updates the main function which checks the Revision tag from
/proc/cpuinfo against known RPi rev numbers. The RPi Zero W and CM3
boards have also been added.

### Implemented caching for board revision lib functions

Using the `use_cached` option, callers of is_model_X functions will
avoid reading and parsing the `/proc/cpuinfo` file each time. This
option was turned on by default as we can safely say the value won't
change during a boot.

As a sidenote, this seems to give a 20x improvement to `is_unsupported_rpi()` from https://github.com/KanoComputing/kano-init/pull/68/files


radujipa:
[(1) As a Pi 1 user, I want to be notified that I'm running on unsupported hardware [0.5]](https://trello.com/c/qGQGUNU4/1729-1-as-a-pi-1-user-i-want-to-be-notified-that-im-running-on-unsupported-hardware-05)

radujipa:
This PR is now ready 🎉

PR #281, opened 2018-02-06
Split kano splash

Ealdwulf:
This PR complements PRs #1 and #2 in the new kano-splash repo.
It removes kano-splash and exports kano_logging as a new library (kano_c_logging, as there is already a kano_logging library exported from kano_qt_sdk).


radujipa:
[(5) As an Engineer I want to simplify the updater by converting all pip packages we use to deb packages](https://trello.com/c/xf06I5vU/1791-5-as-an-engineer-i-want-to-simplify-the-updater-by-converting-all-pip-packages-we-use-to-deb-packages)

PR #287, opened 2018-04-19
Gtk: Fix support for Gtk packages included with Stretch

tombettany:
The upgrade to Stretch brings a new version of Gtk which breaks some of
the CSS. Resolve these issues.

### DIFF
--- a/media/CSS/shutdown.css
+++ b/media/CSS/shutdown.css
@@ -1,14 +1,17 @@
-/*
-# top_bar.css
-#
-# Copyright (C) 2014 Kano Computing Ltd.
-# License: http://www.gnu.org/licenses/gpl-2.0.txt GNU General Public License v2
-#
-#
-*/
+/**
+ * shutdown.css
+ *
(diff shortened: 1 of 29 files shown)


media/CSS/shutdown.css
```diff
@@ -1,14 +1,17 @@
...
+ * Styling for the shutdown
+ */
+
 
 .shutdown_title {
     color: #222222;
-    font: Bariol Bold 16;
+    font-family: Bariol;
+    font-size: 16pt;
+    font-weight: bold;
```
radujipa: Remember to use `bold` with `Bariol` like this:

```
font-family: Bariol bold;
font-size: 16pt;
```

Try to do all like this since a bunch of things will import from this repo.

tombettany: I've made the switch so that everything now uses `font-family: Bariol Bold`.

tombettany:
Superseded by https://github.com/KanoComputing/kano-toolset/pull/290

PR #288, opened 2018-04-25
Adding function to query if running Debian OS is Stretch

skarbat:
For future needs, and QA validation tests


skarbat:
@radujipa FYI: I had to add `ptyhon-sphinx` as a build dependency.
Probably our new buildroot does not have it, just to remember for other doc-able repos.

PR #298, opened 2018-10-05
Fixed caret-color for cursor to be visible in input fields

radujipa:
This should apply the setting globally fixing all Gtk CSS system wide.


Ealdwulf:
Are you sure there are no black input fields?

radujipa:
@Ealdwulf the only one seems to be https://github.com/KanoComputing/make-light/compare/OS-74/fix-gtk-caret-color?expand=1

PR #299, opened 2018-10-08
Fixed traceback caused when calling get_user early during boot

radujipa:
The implementation of the function assumed the existence of LOGNAME
in the os.environ. However, calling this function early during boot
means the variable is not set which caused a traceback. Wrap this
in a safer get and return an empty string as a fallback.

---

```
  File "/usr/bin/kano-sync", line 39, in <module>
    from kano_world.functions import sync, login_using_token, backup_content, \
  File "/usr/lib/python2.7/dist-packages/kano_world/functions.py", line 13, in <module>
    from kano_profile.profile import load_profile, save_profile
  File "/usr/lib/python2.7/dist-packages/kano_profile/profile.py", line 16, in <module>
    from kano_avatar.paths import AVATAR_DEFAULT_LOC
  File "/usr/lib/python2.7/dist-packages/kano_avatar/paths.py", line 13, in <module>
    '/tmp/', get_user(), 'char_gen_scratch', 'character.png')
  File "/usr/lib/python2.7/dist-packages/kano/utils/user.py", line 20, in get_user
    return os.environ['LOGNAME']
  File "/usr/lib/python2.7/UserDict.py", line 40, in __getitem__
    raise KeyError(key)
```

The early call to `kano-sync` seems to be coming from here:
https://github.com/KanoComputing/kano-toolset/blob/master/dhcpcd-hooks/65-kano-dhcp-hook#L40


kano/utils/user.py
```diff
@@ -17,7 +17,7 @@ def get_user_getpass():
 
 
 def get_user():
-    return os.environ['LOGNAME']
+    return os.environ.get('LOGNAME', '')
```
tombettany: Actually, should this return `root` in this case?

radujipa: Not quite sure, seems arbitrary to me as I don't see why not use `getpass.getuser()` for which there is another function... The only place where this seems to be called from is https://github.com/KanoComputing/kano-profile/blob/master/kano_avatar/paths.py#L13

skarbat: I think assuming `root` would be a security harm. I would leave as per Radu's assumption.
If this function fails, surely other things would misbehave in the first place..

PR #303, opened 2019-01-11
[OS-728] Switch to using original WPA supplicant configuration file

skarbat:
This PR changes the wireless network python library to use the original `wpa_supplicant` configuration file.


debian/changelog
```diff
@@ -1,3 +1,9 @@
+kano-toolset (4.2.1-0) unstable; urgency=low
+
+  * Fixed wireless connection on boot for Kano 4.2.0 release
```
radujipa: I'd rewrite this as `Switched to using the original wpa supplicant configuration file`. The version mentioned here is actually 4.2.1 like in the line above.

kano/network.py
```diff
@@ -33,6 +33,7 @@
 DNS_INTERFACES_BACKUP_FILE = '/etc/resolvconf/interface-order.backup'
 SUPPLICANT_CMD = 'wpa_supplicant -D nl80211,wext -t -d -c{} -i{} -f {} -B'
 SUPPLICANT_LOGFILE = '/var/log/kano_wpa.log'
+SUPPLICANT_CONFIG='/etc/wpa_supplicant/wpa_supplicant.conf'
```
radujipa: I'd move some of these to a `paths.py` file.

kano/network.py
```diff
@@ -26,16 +26,9 @@
 import re
 import shutil
 from kano.utils import run_cmd, get_user_unsudoed, run_bg, write_file_contents
+from kano import paths
```
tombettany: Doesn't this need to be

```python
from kano.paths import DNS_FILE, DNS_INTERFACES_FILE, \
    DNS_INTERFACES_BACKUP_FILE, SUPPLICANT_CMD, \
    SUPPLICANT_LOGFILE, SUPPLICANT_CONFIG, \
    INTERNET_UP_FILE, KANO_CONNECT_PIDFILE

Comment (skarbat) [reply kano/network.py]: Tested my changeset on the kit and works just fine. I would prefer to avoid redundancy

**Comment by tombettany** [comment]
Somewhere along the way the tests have failed - let's resolve that. Looks to be from the commit when you moved the path variables out of the file: methinks that a test was looking for them or something.

File: kano/paths.py [hunk 4]
```diff
@@ -28,3 +28,12 @@
 
 common_css_dir = os.path.join(common_media_dir, 'CSS')
 common_images_dir = os.path.join(common_media_dir, 'images')
+
+DNS_FILE = '/etc/resolvconf/resolv.conf.d/base'
+DNS_INTERFACES_FILE = '/etc/resolvconf/interface-order'
+DNS_INTERFACES_BACKUP_FILE = '/etc/resolvconf/interface-order.backup'
+SUPPLICANT_CMD = 'wpa_supplicant -D nl80211,wext -t -d -c{} -i{} -f {} -B'
```
radujipa: Not really a file / path :/

skarbat: Good spot - thanks!

PR #306, opened 2019-07-03
[OS-117] Logging: Prevent tracebacks when argv is not accessible

tombettany:
The logger attempts to read `sys.argv` to determine the executable name
which is doing the logging. There are some ways of launching the logger
(e.g. when using it through one of the C++ Python wrappers) which cause
it to not have this property so the functions traceback. Resolve this by
providing a default value in the event that this isn't set.


radujipa:
This might also solve [OS-117].

